Issue #62, opened 2013-01-17
Change visibility on debug


ebeyrent:
Can you please change $debug in AbstractChannel to be public, or provide a setter so that debug can be enabled and disabled by callers instead of defining a constant?

PR #63, opened 2013-01-17
Enable debug without constants (issue #62)

ebeyrent:
Added support for callers to enable and disable channel debug output
without setting a constant.


postalservice14:
I think this PR can be closed.  Since the `$debug` variable is no longer a boolean that can be set.

videlalvaro:
Yup. I think now the right thing to do would be to introduce a better constructor API for the AMQPConnection* classes that's able to accept a `DebugHelper` class instance.

PR #78, opened 2013-05-03
Null Socket Check

vmadman:
Added a check for a 'null' socket before writing.

### DIFF
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -58,7 +58,10 @@ class StreamIO extends AbstractIO
     {
         $len = strlen($data);
         while (true) {
-            if (false === ($written = fwrite($this->sock, $data))) {
+    		if( $this->sock === null ) {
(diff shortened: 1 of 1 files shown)


vmadman:
Certainly.  As I said, our decision had nothing to do with the library, we had tons of success with it and will probably re-implement it.  We were using php-amqplib to output our application logs to a Rabbit server which was then processed by LogStash in a central location.  

In the end, we found that this approach created an unnecessary point of failure that allowed our logging system to influence our forward performance.  so, now we just write our output to a file and use Logstash's "file" input.  This decouples the application and logging concerns and has proven to be a good bit more resilient, stable, and "rewindable".

Generally we're also making the jump to Node.js and should eventually phase out PHP all together.  Our Node system is using AMQP as it was meant to be used (I think), as a distributed workload orchestrator (i.e. RPC).  Eventually I'm sure we'll need to add our legacy PHP assets into this architecture until the migration to Node.js is complete.. so we'll very likely be using php-amqplib again at that time.

PR #224, opened 2014-11-24
Fixed signals to respond immediately instead of waiting for timeout.

### LINKED ISSUE #146: Resource temporarily unavailable
Hello,

I am trying to publish a message to rabbitmq as follows:
```php
$this->connection = new AMQPConnection('127.0.0.1', '5672', 'myuser','mypassword', 'myvhost');

$this->channel = $this->connection->channel();

$this->message = new AMQPMessage('', array('delivery_mode' => 2));

try {
            $this->message->setBody(json_encode($event));
            $this->channel->basic_publish($this->message, 'amq.direct', 'exchangename');
        }
        catch(Exception $e) {
            var_dump($event);
            print $e->getTraceAsString();
        }

```

Here's the dump of the array i am trying to publish:
```
array(4) {
  ["timestamp"]=>
  string(25) "2014-01-30T09:37:52+00:00"
  ["message"]=>
  string(11) "Rabbit 1764"
  ["priority"]=>
  int(3)
  ["priorityName"]=>
  string(3) "ERR"
}
```

1 message succeeds, 1000 messages succeed. However, going over that number, the library starts firing exceptions as follows:

```
PHP Notice:  fwrite(): send of 134 bytes failed with errno=11 Resource temporarily unavailable in /home/vagrant/Work/Logger/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php on line 65

#0 /home/vagrant/Work/Logger/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php(174): PhpAmqpLib\Wire\IO\StreamIO->write('????????<?(???a...')
#1 /home/vagrant/Work/Logger/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php(203): PhpAmqpLib\Connection\AbstractConnection->write('????????<?(???a...')
#2 /home/vagrant/Work/Logger/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Channel/AMQPChannel.php(803): PhpAmqpLib\Connection\AbstractConnection->send_content(1, 60, 0, 99, '???', '{"timestamp":"2...', Object(PhpAmqpLib\Wire\AMQPWriter))
#3 /home/vagrant/Work/Logger/libs/App/Log/Writer/RabbitMQ.php(62): PhpAmqpLib\Channel\AMQPChannel->basic_publish(Object(PhpAmqpLib\Message\AMQPMessage), 'amq.direct', 'exchangename')
#4 /home/vagrant/Work/Logger/libs/Zend/Log/Writer/Abstract.php(80): App_Log_Writer_RabbitMQ->_write(Array)
#5 /home/vagrant/Work/Logger/libs/Zend/Log.php(157): Zend_Log_Writer_Abstract->write(Array)
#6 /home/vagrant/Work/Logger/testRabbit.php(24): Zend_Log->log('Rabbit 1764', 3)
#7 {main}
```

Please advise.

### LINKED ISSUE #165: PHP Warning when interupting consumer with a signal.
Warning: stream_select(): unable to select [4]: Interrupted system call (max_fd=7) in /Users/stu/Dropbox/im_british_so_i_know_how_to_queue/src/chapter5/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php on line 103

If you add a signal handler, when a signal is caught this warning is output.

I've changed the basic wait loop to look like this:

        while (count($channel->callbacks) && $this->continue) {
            try {
                $channel->wait(null, false, 1);
            } catch (\PhpAmqpLib\Exception\AMQPTimeoutException $e) {
                // Do nothing.
            } catch (\PhpAmqpLib\Exception\AMQPRuntimeException $e) {
                // Do nothing.
            }
        }

The timeout exception is expected, without the timeout the signalhandler never gets called, I guess because the stream stuff lower down is non tickable.

The AMQPRuntimeException is thrown by `AMQPReader::wait` because the call to `select` returns false, I just catch both exceptions & do nothing with them

I wonder if there's a better way to handle this? Before throwing either of the exceptions, check the result of `socket_last_error` and if the result isn't 0 or 10, throw the error? That way you can add @ to mask the warning & still get errors when required.

10 is the error code returned by `socket_last_error` when it is interrupted.

If this sounds like a good idea, I can send you a pull request with the changes.

### LINKED ISSUE #192: Signal triggered shutdown causing AMQPTimeoutException
I want my queue consumers to shut down on a SIGINT/SIGTERM so that they can be controlled using supervisor. I don't want them to shut down while they are processing a message (but they should still shut down if there are queued messages) as most of them work with the database. PR #188 allows for this, however if I attempt to stop my queue and close the channel and connection within the signal handler I get a fatal error:
```
PHP Fatal error:  Uncaught exception 'PhpAmqpLib\Exception\AMQPTimeoutException' with message 'Error sending data. Socket connection timed out' in /home/tadam/dev/amqp_debug/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php:157
Stack trace:
#0 /home/tadam/dev/amqp_debug/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php(390): PhpAmqpLib\Wire\IO\StreamIO->write('????????<?????')
#1 /home/tadam/dev/amqp_debug/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php(501): PhpAmqpLib\Connection\AbstractConnection->write('????????<?????')
#2 /home/tadam/dev/amqp_debug/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Channel/AbstractChannel.php(191): PhpAmqpLib\Connection\AbstractConnection->send_channel_method_frame(1, Array, Object(PhpAmqpLib\Wire\AMQPWriter))
#3 /home/tadam/dev/amqp_debug/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Channel/AMQPChannel.php(927): PhpAmqpLib\Channel\AbstractChannel->send_method_frame(Array, Object(PhpAmqpLib\Wire\AMQPWriter))
#4 /home/tadam/dev/amqp_de in /home/tadam/dev/amqp_debug/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php on line 157
````
I have created a test repo for the issue that can be seen here: https://github.com/TomAdam/php-amqplib-signals-issue. The problem only occurs if the consumer is terminated while there are no messages to process. It has to be down to where the signal is dispatched in #188.

### LINKED ISSUE #235: SSL connections result in  exception "Error Connecting to server(0):"
It seems that AMQPLib is not able to properly negotiate SSL connections. I've tried this both connecting via Stunnel as well as directly via RabbitMQs built in SSL support. Each time the socket is closed abruptly by AMQPLib and I receive the following exception:

"Error Connecting to server(0): "

```php
try {
        $ssl_options = array(
                'CAfile' => 'cacert.crt',
                'certfile' => 'app-cert.pem',
                'keyfile' => 'app-key.pem',
        );
        $options = array(
                'connection_timeout' => 10,
        );

        $connection = new AMQPSSLConnection('---', 5671, '---', '---', '/deployment', $ssl_options, $options);
} catch (Exception $e) {
        echo $e->getMessage();
}
```

Using the exact same certificates and keys I'm able to negotiate SSL correctly with openssl s_client as well as with the Python pika library:
```python
import json
import sys
import urllib

# set up the connection
connection = pika.BlockingConnection(
   pika.ConnectionParameters( 
      host='---', port=5671, 
     credentials=pika.PlainCredentials('---', '---'), 
     virtual_host='/deployment',   
     socket_timeout=3, 
     ssl_options={'ca_certs': 'cacert.crt', 'certfile': 'app-cert.pem', 'keyfile': 'app-key.pem'}, ssl=True))
```

roihq:
Streams that are blocking are unable to interpret signals until a timeout occurs. When working on a connection with a higher timeout or using heartbeats to keep the connection alive, signals will not work. 

Workarounds have been mentioned which set a timeout and loop every x seconds to do a check but this is inefficient for the cpu. Using stream_select with a value of null is the easiest on resources while waiting for input. The only way that input will be caught is if the stream is not blocking.

Fixes #165
Fixes #192 

With a stressed connection and also when writing larger values to the stream the os can queue writes and result in hidden issues with the stream. I have removed the @ and have added a few allowed errors that can occur. Everything else will be raised as an exception so it will be easier for people to debug connection issues vs library issues.

Fixes #146 
Fixes #235

### DIFF
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -844,12 +844,18 @@ class AMQPChannel extends AbstractChannel
      *
      * @param string $consumer_tag
      * @param bool $nowait
+     * @param bool $noreturn
      * @return mixed
      */
-    public function basic_cancel($consumer_tag, $nowait = false)
+    public function basic_cancel($consumer_tag, $nowait = false, $noreturn = false)
     {
         list($class_id, $method_id, $args) = $this->protocolWriter->basicCancel($consumer_tag, $nowait);
         $this->send_method_frame(array($class_id, $method_id), $args);
+
+        if ($nowait || $noreturn) {
+            unset($this->callbacks[$consumer_tag]);
+            return $consumer_tag;
+        }
 
         return $this->wait(array(
             $this->waitHelper->get_wait('basic.cancel_ok')
@@ -876,6 +882,7 @@ class AMQPChannel extends AbstractChannel
     {
         $consumer_tag = $args->read_shortstr();
         unset($this->callbacks[$consumer_tag]);
+        return $consumer_tag;
     }
 
     /**
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -9,7 +9,10 @@ use PhpAmqpLib\Wire\AMQPWriter;
 
 class StreamIO extends AbstractIO
 {
-    /** @var  string */
+    /** @var string */
+    protected $protocol;
+
+    /** @var string */
     protected $host;
 
     /** @var int */
@@ -36,8 +39,14 @@ class StreamIO extends AbstractIO
     /** @var float */
     protected $last_write;
 
+    /** @var array */
+    protected $last_error;
+
     /** @var resource */
     private $sock;
+
+    /** @var bool */
+    private $canSelectNull;
 
     /** @var bool */
     private $canDispatchPcntlSignal;
@@ -60,6 +69,7 @@ class StreamIO extends AbstractIO
         $keepalive = false,
         $heartbeat = 0
     ) {
+        $this->protocol = 'tcp';
         $this->host = $host;
         $this->port = $port;
         $this->connection_timeout = $connection_timeout;
@@ -67,8 +77,20 @@ class StreamIO extends AbstractIO
         $this->context = $context;
         $this->keepalive = $keepalive;
         $this->heartbeat = $heartbeat;
-        $this->canDispatchPcntlSignal = extension_loaded('pcntl') && function_exists('pcntl_signal_dispatch')
+        $this->canSelectNull = true;
+        $this->canDispatchPcntlSignal = extension_loaded('pcntl') 
+            && function_exists('pcntl_signal_dispatch')
             && (defined('AMQP_WITHOUT_SIGNALS') ? !AMQP_WITHOUT_SIGNALS : true);
+
+        if (is_null($this->context)) {
+            $this->context = stream_context_create();
+        } else {
+            $this->protocol = 'ssl';
+            // php bugs 41631 & 65137 prevent select null from working on ssl streams
+            if (PHP_VERSION_ID < 50436) {
+                $this->canSelectNull = false;
+            }
+        }
     }
 
     /**
@@ -81,33 +103,44 @@ class StreamIO extends AbstractIO
     {
         $errstr = $errno = null;
 
-        if ($this->context) {
-            $remote = sprintf('ssl://%s:%s', $this->host, $this->port);
-            $this->sock = @stream_socket_client(
-                $remote,
-                $errno,
-                $errstr,
-                $this->connection_timeout,
-                STREAM_CLIENT_CONNECT,
-                $this->context
-            );
-        } else {
-            $remote = sprintf('tcp://%s:%s', $this->host, $this->port);
-            $this->sock = @stream_socket_client(
-                $remote,
-                $errno,
-                $errstr,
-                $this->connection_timeout,
-                STREAM_CLIENT_CONNECT
+        $remote = sprintf(
+            '%s://%s:%s', 
+            $this->protocol,
+            $this->host,
+            $this->port
+        );
+
+        set_error_handler(array($this, 'error_handler'));
+        
+        $this->sock = stream_socket_client(
+            $remote,
+            $errno,
+            $errstr,
+            $this->connection_timeout,
+            STREAM_CLIENT_CONNECT,
+            $this->context
+        );
+        
+        restore_error_handler();
+
+        if (false === $this->sock) {
+            throw new AMQPRuntimeException(
+                sprintf(
+                    'Error Connecting to server(%s): %s ', 
+                    $errno,
+                    $errstr
+                ),
+                $errno
             );
         }
-
-        if (!$this->sock) {
-            throw new AMQPRuntimeException(sprintf(
-                'Error Connecting to server (%s): %s',
-                $errno,
-                $errstr
-            ), $errno);
+        
+        if (false === stream_socket_get_name($this->sock, true)) {
+            throw new AMQPRuntimeException(
+                sprintf(
+                    'Connection refused: %s ', 
+                    $remote
+                )
+            );
         }
 
         list($sec, $uSec) = MiscHelper::splitSecondsMicroseconds($this->read_write_timeout);
@@ -115,7 +148,16 @@ class StreamIO extends AbstractIO
             throw new AMQPIOException('Timeout could not be set');
         }
 
-        stream_set_blocking($this->sock, 1);
+        // php cannot capture signals while streams are blocking
+        if ($this->canDispatchPcntlSignal) {
+            stream_set_blocking($this->sock, 0);
+            stream_set_write_buffer($this->sock, 0);
+            if (function_exists('stream_set_read_buffer')) {
+                stream_set_read_buffer($this->sock, 0);
+            }
+        } else {
+            stream_set_blocking($this->sock, 1);
+        }
 
         if ($this->keepalive) {
             $this->enable_keepalive();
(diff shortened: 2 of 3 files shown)


PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -276,7 +285,7 @@ public function select($sec, $usec)
         $write = null;
         $except = null;
 
-        return stream_select($read, $write, $except, $sec, $usec);
+        return @stream_select($read, $write, $except, $sec, $usec);
```
roihq: Updated code and removed @

Makefile
```diff
@@ -1,5 +1,5 @@
 test:
-	vendor/bin/phpunit
+	vendor/bin/phpunit --debug
```
roihq: Removed

PhpAmqpLib/Channel/AMQPChannel.php
```diff
@@ -844,13 +844,19 @@ public function basic_nack($delivery_tag, $multiple = false, $requeue = false)
      *
      * @param string $consumer_tag
      * @param bool $nowait
+     * @param bool $noreturn
      * @return mixed
      */
-    public function basic_cancel($consumer_tag, $nowait = false)
+    public function basic_cancel($consumer_tag, $nowait = false, $noreturn = false)
```
videlalvaro: As I said earlier, `$noreturn` is not part of the spec, and it seems to duplicate what `$nowait` is made for

roihq: I will try and make a demo when I get a chance. I'll remove it if it moves the pull request along. I assumed the spec specific functionality was in the Protocol folder. There seemed to be a few things already in here that didn't comply. 

No wait doesn't expect a message response from the broker and no return doesn't wait for one. It's different in that I still need a response from the broker but I do not want to wait (block) in the code for one. Getting signals to work in this pull request uncovered the issue. When I send it a signal to stop consuming and close the code wouldn't exit. Signals can be sent at any point and after dealing with them it returns back to the last place it left off.

PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -176,17 +238,24 @@ public function read($n)
...
+        
+        while ($written < $len) {
+
+            if (!is_resource($this->sock)) {
                 throw new AMQPRuntimeException('Broken pipe or closed connection');
             }
 
-            if (false === ($written = @fwrite($this->sock, $data))) {
+            set_error_handler(array($this, 'error_handler'));
+            $buffer = fwrite($this->sock, $data);
```
rhysr: Should the fwrite() be wrapped in a try catch?
error_hander() can throw an exception and if it does, restore_error_handler() won't get called.

PhpAmqpLib/Channel/AMQPChannel.php
```diff
@@ -844,13 +844,19 @@ public function basic_nack($delivery_tag, $multiple = false, $requeue = false)
...
     {
         list($class_id, $method_id, $args) = $this->protocolWriter->basicCancel($consumer_tag, $nowait);
         $this->send_method_frame(array($class_id, $method_id), $args);
 
+        if ($nowait || $noreturn) {
```
videlalvaro: I still don't get why you need the `$noreturn`. Just set the `$nowait` var to `true` and it will do the same. What am I missing?

PhpAmqpLib/Channel/AMQPChannel.php
videlalvaro: when sending a `basic.cancel` the server can still send messages to the consumer until it receives the actual `basic.cancel`. Wouldn't this allow for a race condition where messages could be lost?

videlalvaro:
ping @roihq I really want to get this merged. Would you mind explaining why the need for the new TCP/SSL handling in the StreamIO class?

PhpAmqpLib/Channel/AMQPChannel.php
roihq: I have added a demo for you to try. If you can find another way around it I will be happy to change my code.

roihq:
@videlalvaro I have re-based my changes with the latest code. The current TCP/SSL is broken as mentioned above. https://github.com/reactphp/react helped in finding solutions to the issues. I would recommend looking at using the event-loop or stream code form that library if you want to separate amqplib form dealing with stream/event code in php.

PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -132,40 +174,60 @@ public function reconnect()
...
         }
 
-        if (mb_strlen($res, 'ASCII') != $n) {
-            throw new AMQPIOException(sprintf(
-                'Error reading data. Received %s instead of expected %s bytes',
-                mb_strlen($res, 'ASCII'),
-                $n
-            ));
+        if (mb_strlen($data, 'ASCII') !== $len) {
+            throw new AMQPRuntimeException(
```
videlalvaro: Any reason why you changed the Exception type? If you have to change it, I think you need to update the PHP Doc section of this function

videlalvaro:
@roihq what do you think about the Exception Type comment above?

michaelklishin:
@roihq any updates on the comment above?

roihq:
@videlalvaro @michaelklishin I'm sorry I'm away for awhile. If someone can take this over it would be appreciated.

plouc:
@videlalvaro, any chance to see those fixes released soon ?

videlalvaro:
@plouc I'll try my best but I'm quite busy at work.

plouc:
@videlalvaro thx a lot, I understand it :)

PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -194,14 +263,42 @@ public function write($data)
...
+        }
+
+        // stream_select warning that it has been interrupted by a signal
+        if (strstr($errstr, 'Interrupted system call')) {
+             // it's allowed while processing signals
+             return;
+        }
+
+        // raise all other issues to exceptions
+        throw new \ErrorException($errstr, 0, $errno, $errfile, $errline);
```
lsmith77: it seems like this change causes triggering of a different exception:
https://travis-ci.org/zendframework/ZendDiagnostics/jobs/82909734#L377

the test code is here:
https://github.com/zendframework/ZendDiagnostics/blob/master/tests/ZendDiagnosticsTest/ChecksTest.php#L55

maybe this should at least be mentioned in the CHANGELOG file

videlalvaro: You are right, sorry about this. I've added a note here: https://github.com/videlalvaro/php-amqplib/blob/master/CHANGELOG.md

PR #236, opened 2015-01-12
Incorrect quoting of some sprintf formats.

fredericmora:
The format argument of sprintf should be bracketed in double quote if interpolation is  needed.
 The $x is supposed to expand to either x or X. The single quotes prevent this interpolation and instead are seen as a positional argument spec n$, which is incorrect here.

As it stands, the code generates a "sprintf(): Too few arguments : warning.

### DIFF
--- a/PhpAmqpLib/Helper/MiscHelper.php
+++ b/PhpAmqpLib/Helper/MiscHelper.php
@@ -77,7 +77,7 @@ class MiscHelper
         // Iterate string
         for ($i = $j = 0; $i < $len; $i++) {
             // Convert to hexidecimal
-            $hexi .= sprintf('%02$x ', ord($data[$i]));
+            $hexi .= sprintf("%02$x ", ord($data[$i]));
 
             // Replace non-viewable bytes with '.'
             if (ord($data[$i]) >= 32) {
@@ -97,7 +97,7 @@ class MiscHelper
             // Add row
             if (++$j === 16 || $i === $len - 1) {
                 // Join the hexi / ascii output
-                $dump .= sprintf('%04$x  %-49s  %s', $offset, $hexi, $ascii);
+                $dump .= sprintf("%04$x  %-49s  %s", $offset, $hexi, $ascii);
 
                 // Reset vars
                 $hexi = $ascii = '';
--- a/tests/Unit/Helper/MiscHelperTest.php
+++ b/tests/Unit/Helper/MiscHelperTest.php
@@ -33,4 +33,12 @@ class MiscHelperTest extends \PHPUnit_Framework_TestCase
             array('3.123456', array(3, 123456)),
         );
     }
+
+    public function testHexDump()
+    {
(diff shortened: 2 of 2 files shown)


PhpAmqpLib/Helper/MiscHelper.php
```diff
@@ -77,7 +77,7 @@ public static function hexdump($data, $htmloutput = true, $uppercase = false, $r
         // Iterate string
         for ($i = $j = 0; $i < $len; $i++) {
             // Convert to hexidecimal
-            $hexi .= sprintf('%02$x ', ord($data[$i]));
+            $hexi .= sprintf("%02$x ", ord($data[$i]));
```
fprochazka: How do you know if this wasn't itentional? Would you mind creating a test case, so if this really is a bug, nobody will break it again?

fprochazka:
Would you mind creating a test case, so if this really is a bug, nobody will break it again?

fredericmora:
Added test case.

tests/Unit/Helper/MiscHelperTest.php
```diff
@@ -33,4 +33,12 @@ public function getInputOutputForSplitSecondsMicroseconds()
             array('3.123456', array(3, 123456)),
         );
     }
+
+    public function testHexDump()
+    {
+        $res = MiscHelper::hexdump('FM', 'hmtlOutput' == false, 'uppercase' == false, 'return' == true);
```
fprochazka: Are you sure about this? It's really confusing. Wouldn't this be better?

```php
 $res = MiscHelper::hexdump('FM', $htmlOutput = false, $uppercase = false, $return = true);
```

fredericmora: Sure, you can use your method because it's test code. It has the side effect of setting the $htmlOutput, $uppercase and $return variable, though, which is why it's not a good practice in the general case.

The method I use is only valid for "documenting" boolean arguments passed to external methods. 'x'==false evaluates to false, 'x'==true evals to true, for any string. Since it doesn't work for numerical arguments, you can argue that it's not very generalizable either...

fprochazka: Yeah, I know that it works, but It's super confusing and OS libs should IMHO contain as little as possible hacks of this type.

fprochazka: FYI, I'm not a maintainer of this lib, only a user :)

fredericmora: Awright, awright :-)

I'm going to commit a non-hack, cleaned-up call. Give me 10 mins.

fprochazka: I'm not a maintainer, just saying what I believe is the best. Take as much time as you need, @videlalvaro is probably also going to take some time :) 

And thank you for fixing this!

PR #237, opened 2015-01-14
Stick to single quote convention + minor changes

eexit:
Thank you very much for the fix on #236, really and thank you more the reactivity.

This small PR brings an alternative to single quote "convention" break and bring back the single quote. I also made minor language-wise changes.

### DIFF
--- a/PhpAmqpLib/Helper/MiscHelper.php
+++ b/PhpAmqpLib/Helper/MiscHelper.php
@@ -67,23 +67,23 @@ class MiscHelper
         // Init
         $hexi = '';
         $ascii = '';
-        $dump = ($htmloutput === true) ? '<pre>' : '';
+        $dump = $htmloutput ? '<pre>' : '';
         $offset = 0;
         $len = mb_strlen($data, 'ASCII');
 
         // Upper or lower case hexidecimal
-        $x = ($uppercase === false) ? 'x' : 'X';
+        $hexFormat = $uppercase ? 'X' : 'x';
 
         // Iterate string
         for ($i = $j = 0; $i < $len; $i++) {
             // Convert to hexidecimal
-            $hexi .= sprintf("%02$x ", ord($data[$i]));
+            // We must use concatenation here because the $hexFormat value
+            // is needed for sprintf() to parse the format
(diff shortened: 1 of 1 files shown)


PhpAmqpLib/Helper/MiscHelper.php
```diff
@@ -67,23 +67,23 @@ public static function hexdump($data, $htmloutput = true, $uppercase = false, $r
...
-        $x = ($uppercase === false) ? 'x' : 'X';
+        $hexFormat = $uppercase ? 'X' : 'x';
 
         // Iterate string
         for ($i = $j = 0; $i < $len; $i++) {
             // Convert to hexidecimal
-            $hexi .= sprintf("%02$x ", ord($data[$i]));
+            // We must use concatenation here because the $hexFormat value
+            // is needed for sprintf() to parse the format
+            $hexi .= sprintf('%02' .  $hexFormat . ' ', ord($data[$i]));
```
videlalvaro: why not sprintf the `$hexFormat` variable as well?

PhpAmqpLib/Helper/MiscHelper.php
```diff
@@ -97,7 +97,9 @@ public static function hexdump($data, $htmloutput = true, $uppercase = false, $r
             // Add row
             if (++$j === 16 || $i === $len - 1) {
                 // Join the hexi / ascii output
-                $dump .= sprintf("%04$x  %-49s  %s", $offset, $hexi, $ascii);
+                // We must use concatenation here because the $hexFormat value
+                // is needed for sprintf() to parse the format
+                $dump .= sprintf('%04' . $hexFormat . '  %-49s  %s', $offset, $hexi, $ascii);
```
videlalvaro: same as above

PhpAmqpLib/Helper/MiscHelper.php
fprochazka: because the variable contains the modifier type that is used for sprintf formating

videlalvaro: Oh… stupid me

PR #245, opened 2015-02-20
Declared the class AbstractChannel as being an abstract class

theTeddyBear:
I've declared the class AbstractChannel as being an abstract class, as the naming suggests


videlalvaro:
What's the point of this change? Does it break backwards compatibility?

theTeddyBear:
It does not break backwards compatibility as far as I can see - it didn't in any testing I did.

The point of this change is to not incorrectly identify a class as being Abstract when it has not been declared as being so.

PR #252, opened 2015-03-03
add grongor's fix (adapted) for UNEXPECTED_FRAME

amne:


videlalvaro:
Could you explain what are you trying to do here?

videlalvaro:
I merged https://github.com/videlalvaro/php-amqplib/pull/298 instead, since that one has tests as well. Thanks

PR #254, opened 2015-03-07
Fix for unknown delivery tag

liborm85:
This PR fixed randomly occurring error "'Server nack\'ed unknown delivery_tag 1.00000000" and problems.

This resolves issue: https://github.com/videlalvaro/php-amqplib/issues/180


videlalvaro:
Thanks!

PR #262, opened 2015-04-01
Include PHP 7 and HHVM Nightly builds in testing suite

frankmayer:
Since HHVM tests already pass, it would be good to start testing the upcoming versions of PHP & HHVM. Of course we're allowing failures for those cases.


frankmayer:
And they do pass!! :smile:

PR #264, opened 2015-04-06
StreamIO::select() fix warning when catching a signal during stream_select

holycheater:
Similar to problem in reactphp:
https://github.com/reactphp/react/issues/301
https://github.com/reactphp/react/issues/296
https://github.com/reactphp/react/pull/297

http://php.net/stream_select
On error FALSE is returned and a warning raised (this can happen if the system call is interrupted by an incoming signal).

### DIFF
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -276,7 +276,18 @@ class StreamIO extends AbstractIO
         $write = null;
         $except = null;
 
-        return stream_select($read, $write, $except, $sec, $usec);
+        $result = @stream_select($read, $write, $except, $sec, $usec);
+        /* stream_select return false when we catch a signal with pcntl_signal
+         * and generates a warning. Trying to avoid that here
+         */
+        if ($result === false) {
+            $err = error_get_last();
+            if (isset($err['message']) && stripos($err['message'], 'interrupted system call')) {
+                return stream_select($read, $write, $except, $sec, $usec);
+            }
+            return false;
+        }
+        return $result;
     }
 
     /**


PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -276,7 +276,18 @@ public function select($sec, $usec)
         $write = null;
         $except = null;
 
-        return stream_select($read, $write, $except, $sec, $usec);
+        $result = @stream_select($read, $write, $except, $sec, $usec);
```
videlalvaro: is it required to used the `@` operator?

roihq: Please see #224 this has been fix without needing @ as requested by @fprochazka

holycheater: The problem: stream_select generates a warning when app is interrupted by a signal.
So if you have a custom error handler that treats warnings as fatal then app crashes.

kler:
I suggest to do like you did with `fwrite`calls, add `@` and wrap in an exception.

PhpAmqpLib/Wire/IO/StreamIO.php
roihq: If you look at the error handler it ignores the case you are talking about. The pull request is for fixing signals and the first few changes that I did using @ exactly like you have were not used.

videlalvaro:
Would you mind bringing this PR up to date with master?

postalservice14:
Looks like this can be closed.  It appears to be fixed by the error_handler: https://github.com/videlalvaro/php-amqplib/blob/master/PhpAmqpLib/Wire/IO/StreamIO.php#L376

videlalvaro:
@holycheater closing as per @postalservice14 comment. If you still feel the need for this PR, then please re-open it.

PR #267, opened 2015-04-13
Allow a max body size to be set per Channel

IMSoP:
If a very large message is received by a consumer, PHP's memory limit may be reached
before the basic_consume callback is even called. To avoid this, the Channel needs to avoid constructing the Message.

Proposed solution is a new public function setMaxBodySize(int $bytes) on Channel; body sizes exceeding this will be truncated, and $msg->is_truncated flag set. A new public property $msg->body_size reflects the true body size (i.e. it higher than strlen($msg->body)).

This allows the callback to check the flat and decide whether to use the truncated data, or nack the message, etc.

The default remains unlimited to avoid compatibility issues, although a default value based on `ini_set('memory_limit')` might be possible.

### DIFF
--- a/PhpAmqpLib/Channel/AbstractChannel.php
+++ b/PhpAmqpLib/Channel/AbstractChannel.php
@@ -42,6 +42,11 @@ class AbstractChannel
 
     /** @var string */
     protected $protocolVersion;
+    
+    /**
+     * @var int
+     */
+    protected $body_size_max;
 
     /** @var \PhpAmqpLib\Helper\Protocol\Protocol080|\PhpAmqpLib\Helper\Protocol\Protocol091 */
     protected $protocolWriter;
@@ -128,6 +133,18 @@ class AbstractChannel
     public function getChannelId()
     {
         return $this->channel_id;
+    }
+    
+    public function setBodySizeLimit($max_bytes)
+    {
+        $max_bytes = intval($max_bytes);
+        
+        if ( $max_bytes > 0 ) {
+            $this->body_size_max = $max_bytes;
+        }
+        else {
+            unset($this->body_size_max);
+        }
     }
 
     /**
@@ -247,7 +264,8 @@ class AbstractChannel
 
         $msg = new AMQPMessage();
         $msg->load_properties($this->msg_property_reader);
-
+        $msg->body_size = $body_size;
+        
         $body_parts = array();
         $body_received = 0;
         while (bccomp($body_size, $body_received, 0) == 1) {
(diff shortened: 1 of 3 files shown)


videlalvaro:
Can you document this feature on the README?

IMSoP:
Good point. I've added a quick explanation.

videlalvaro:
What happens to the rest of the message body, how do I retrieve it?

IMSoP:
Currently, the rest of the message is just discarded - the frames are read from the channel and ignored by the `continue` statement. I suppose you could theoretically block the channel and give the callback some kind of stream to consume the remaining content, but that's really a whole new API, and this was intended purely as a protection against crashes.

In my use case, these were logging events being pushed into a database, so a streaming API wouldn't have been particularly useful - ultimately, I want to either discard or truncate these oversized events anyway.

If there is a possibility of another consumer being able to handle larger payloads on the same queue, you can nack() the message, and configure an appropriate Dead Letter Exchange. I considered doing this automatically, but figured that it made more sense for the consumer to decide what to do.

videlalvaro:
Mind adding an explanation to the docs you added explicitly saying that the rest of the body is discarded + your nack() suggestion. Then I can hit the green button

PhpAmqpLib/Channel/AbstractChannel.php
```diff
@@ -129,6 +134,18 @@ public function getChannelId()
...
+    
+    public function setBodySizeLimit($max_bytes)
+    {
+        $max_bytes = intval($max_bytes);
+        
+        if ( $max_bytes > 0 ) {
+            $this->body_size_max = $max_bytes;
+        }
+        else {
+            unset($this->body_size_max);
```
fprochazka: please, never unset properties! Just set them to NULL

PhpAmqpLib/Channel/AbstractChannel.php
```diff
@@ -129,6 +134,18 @@ public function getChannelId()
...
     }
+    
+    public function setBodySizeLimit($max_bytes)
+    {
+        $max_bytes = intval($max_bytes);
+        
+        if ( $max_bytes > 0 ) {
+            $this->body_size_max = $max_bytes;
+        }
+        else {
```
fprochazka: also, coding standard

PhpAmqpLib/Channel/AbstractChannel.php
IMSoP: Good point. Fixed.

PhpAmqpLib/Channel/AbstractChannel.php
IMSoP: Do you mean the else shouldn't be preceded by a newline? If so, also fixed; if not, please be more specific.

IMSoP:
Apologies for the force push, I wanted to fix the author on the last few commits.

videlalvaro:
@fprochazka any more comments on code style? Otherwise I'll merge this

IMSoP:
Bump :)

PR #275, opened 2015-05-20
Update demo config

sergeyklay:


videlalvaro:
why do we need this change? If you get the source from github and then run `composer install` everything works as expected

sergeyklay:
I proposed these changes, because only in this way it is possible to use these examples.

videlalvaro:
If you clone the library directly from the Github repo, then run composer install, the examples work perfectly. They might not work if `php-amqplib` is a dependency of one of your repos or something like that. The path you provide in the changes suggest you have a different directory structure. I have something like:

/PhpAmqpLib
/demo
/vendor

sergeyklay:
Thank you for the explanation

PR #298, opened 2015-07-23
Prevent UNEXPECTED_FRAME exception

kgorski:
Prevent "UNEXPECTED_FRAME - expected method frame, got non method frame instead" exception when message body length is equal multiplication of 131064 bytes


michaelklishin:
Sounds like a good candidate for a test case (or two).

videlalvaro:
Seems to be a duplicate of https://github.com/videlalvaro/php-amqplib/pull/252 but since you have tests I prefer to merge this one :+1:

PR #301, opened 2015-07-29
Fixed fatal error in some cases on Channel.wait with timeout

todoshcenko:
Changed demo/amqp_consumer.php
```php
while (true)
{
    try
    {
        if (count($ch->callbacks))
        {
            $ch->wait(null, false, 1);
        }
    }
    catch (AMQPTimeoutException $e)
    {
    }
    catch (AMQPProtocolConnectionException $e)
    {
    }
}
```

If RabbitMQ restarts while this code is running it leads to PHP fatal error "Call to a member function getTimeout()"


todoshcenko:
This code only demonstrates the problem. I think a library shouldn't allow fatal errors to occur

videlalvaro:
Thanks

PR #303, opened 2015-08-05
If AMQPConnection is deprecated; why extending it?

frederikbosch:


videlalvaro:
Perhaps for backwards compatibility, in the case someone was type hinting their functions?

frederikbosch:
@videlalvaro Yeah, I understand that. It is deprecated for over a year now, and should not be just anymore since [Feb 1, 2013](https://github.com/videlalvaro/php-amqplib/commit/68a26730b46c9a87c1c76c297bc2c752f736be0e). Moreover, I just find it weird that only the `AMQPLazyConnection` is extending the `AMQPConnection`, where as the `AMQPStreamConnection` is extending the `AbstractConnection`.

In my case: I am using the lazy connection, but my IDE indicates that I am using deprecated classes. And that is not the case, right? Therefore it would be my suggestion to drop the extension.

However, I am now noticing that all your examples still contain the `AMQPConnection`. Since that is the case: we could better update those examples, then merge this in 3.0.

frederikbosch:
Oh, you just merged it.

frederikbosch:
Since your examples contain `AMQPConnection`, I would suggest the following.

```php
class AMQPLazyConnection extends AMQPStreamConnection {}
class AMQPConnection extends AMQPLazyConnection {}
```

Then user of `AMQPLazyConnection` will not see deprecated warning. And you will not break code of people that are still using `AMQPConnection`.

frederikbosch:
Just deleted my previous comment, not a great idea. Updating the examples would be a great idea.

videlalvaro:
I merged it since I plan to bump the library's minor version, so if someone
is still type hinting on the deprecated class they would have to update
their code. About the examples, feel free to send a new PR now that I'm
merge-trigger-happy ;-)

PR #305, opened 2015-08-08
Fix of wrong typehinting of password parameter

DawidMazurek:
Wrong typehinting introduced some unclarity, beacase password was typehinted as bool. Now it is changed to string.


videlalvaro:
thanks

PR #307, opened 2015-09-01
Add .gitattributes to slim down composer packages.

jdufresne:
When composer is used as a dependency manager for large web apps,
transferring lots of extra files (such as tests and documentation) can
be very time and bandwidth consuming. By using gitattributes we can
tell composer what files to ignore when distributing through composer.


videlalvaro:
Thanks

PR #319, opened 2015-09-28
Fix not restoring error handler

PepijnSenders:
By overwriting the error handler there would be an infinite loop of exceptions, one should always try to restore the error handler. This is just an example of how it should be done througout the project.

Note: one can also use `finally` if PHP >=5.5.

### DIFF
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -112,20 +112,14 @@ class StreamIO extends AbstractIO
 
         set_error_handler(array($this, 'error_handler'));
 
-        try {
-            $this->sock = stream_socket_client(
-                $remote,
-                $errno,
-                $errstr,
-                $this->connection_timeout,
-                STREAM_CLIENT_CONNECT,
-                $this->context
-            );
-        } catch (\Exception $e) {
-            restore_error_handler();
-
-            throw $e;
-        }
+        $this->sock = stream_socket_client(
+            $remote,
+            $errno,
+            $errstr,
+            $this->connection_timeout,
+            STREAM_CLIENT_CONNECT,
+            $this->context
+        );
 
         restore_error_handler();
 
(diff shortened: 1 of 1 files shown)


PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -104,40 +104,46 @@ public function connect()
...
-            $errno,
-            $errstr,
-            $this->connection_timeout,
-            STREAM_CLIENT_CONNECT,
-            $this->context
-        );
-        
+
+        try {
+            $this->sock = stream_socket_client(
```
postalservice14: I don't believe `stream_socket_client` throws exceptions.  Am I wrong?

PepijnSenders: It is going to throw an exception by the error handler that is being set by line 113.

postalservice14: Ahh gotcha, makes sense.  Sorry didn't catch that.

postalservice14:
:+1:

PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -104,40 +104,46 @@ public function connect()
...
+                $remote,
+                $errno,
+                $errstr,
+                $this->connection_timeout,
+                STREAM_CLIENT_CONNECT,
+                $this->context
+            );
+        } catch (\Exception $e) {
+            restore_error_handler();
```
videlalvaro: Why not restore the error handler here: https://github.com/videlalvaro/php-amqplib/blob/master/PhpAmqpLib/Wire/IO/StreamIO.php#L300 before throwing the ErrorException then?

PepijnSenders: That's actually a good point :+1:

videlalvaro: Would you mind sending a new PR? I think then the try/catch introduced here is also not required

PepijnSenders: Sure!

videlalvaro:
Thanks

PR #201, opened 2014-09-21
Minor complexity optimizations + code organization

eexit:
According to #193, here's the PR that refacto some parts of the lib: condition simplifications, better variable naming, better exception message building, etc.

Feedback and contribution is very welcomed.

Thanks,

PR #323, opened 2015-10-05
Minor complexity optimizations + code organization

postalservice14:
Ported changes from PR #201.  All credit to @eexit

### DIFF
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -192,9 +192,9 @@ class AMQPChannel extends AbstractChannel
      * Confirm a channel close
      * Alias of AMQPChannel::do_close()
      *
-     * @param AMQPReader $args
+     * @param AMQPReader $reader
      */
-    protected function channel_close_ok($args)
+    protected function channel_close_ok(AMQPReader $reader)
     {
         $this->do_close();
     }
@@ -237,7 +237,7 @@ class AMQPChannel extends AbstractChannel
      * @param AMQPReader $args
      * @return bool
      */
-    protected function channel_flow_ok($args)
+    protected function channel_flow_ok(AMQPReader $args)
     {
         return $args->read_bit();
     }
@@ -261,9 +261,9 @@ class AMQPChannel extends AbstractChannel
     }
 
     /**
-     * @param AMQPReader $args
+     * @param AMQPReader $reader
      */
-    protected function channel_open_ok($args)
+    protected function channel_open_ok(AMQPReader $reader)
     {
         $this->is_open = true;
 
@@ -308,12 +308,12 @@ class AMQPChannel extends AbstractChannel
     /**
      * Grants access to server resources
      *
-     * @param AMQPReader $args
+     * @param AMQPReader $reader
      * @return string
      */
-    protected function access_request_ok($args)
+    protected function access_request_ok($reader)
     {
-        $this->default_ticket = $args->read_short();
+        $this->default_ticket = $reader->read_short();
 
         return $this->default_ticket;
     }
@@ -369,8 +369,9 @@ class AMQPChannel extends AbstractChannel
 
     /**
      * Confirms an exchange declaration
+     * @param AMQPReader $reader
      */
-    protected function exchange_declare_ok($args)
+    protected function exchange_declare_ok(AMQPReader $reader)
     {
     }
 
@@ -410,8 +411,10 @@ class AMQPChannel extends AbstractChannel
 
     /**
      * Confirms deletion of an exchange
+     *
+     * @param AMQPReader $reader
      */
-    protected function exchange_delete_ok($args)
+    protected function exchange_delete_ok(AMQPReader $reader)
     {
     }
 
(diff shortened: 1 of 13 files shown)


PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -380,9 +380,7 @@ public function send_content($channel, $class_id, $weight, $body_size, $packed_p
      */
     public function prepare_content($channel, $class_id, $weight, $body_size, $packed_properties, $body, $pkt = null)
     {
-        if (empty($pkt)) {
-            $pkt = new AMQPWriter();
-        }
+        $pkt = $pkt ?: new AMQPWriter();
```
videlalvaro: Since when is `?:` supported?

PhpAmqpLib/Message/AMQPMessage.php
```diff
@@ -56,5 +53,7 @@ public function __construct($body = '', $properties = null)
     public function setBody($body)
     {
         $this->body = $body;
+
+        return $this;
```
videlalvaro: Is there a need for a "fluent API" here?

PhpAmqpLib/Wire/GenericContent.php
```diff
@@ -126,7 +122,7 @@ public function load_properties($r)
         while (true) {
             $flag_bits = $r->read_short();
             $flags[] = $flag_bits;
-            if (($flag_bits & 1) == 0) {
+            if (($flag_bits & 1) === 0) {
```
videlalvaro: could you explain under which circumstances this code requires the `===` instead of `==`. Same comment for the change below

videlalvaro:
Left a few comments, once they are addressed, we can merge this. Thanks

PhpAmqpLib/Connection/AbstractConnection.php
postalservice14: Since 5.3

PhpAmqpLib/Message/AMQPMessage.php
postalservice14: The phpdocs said it returned $this.  So I just made it true.  Personally I like fluent API for setters.  But that's just my opinion.

videlalvaro: Cool. I think it's fine to have it, just that the parent class doesn't implement this kind of API

PhpAmqpLib/Wire/GenericContent.php
postalservice14: `===` enforces type-checking in the comparison.  So `"0" === 0` returns false.  I think it's always good to do when able.  Makes your code more stable.   PHP's loosly typed-ness is nice in that it's more forgiving and a great feature, but can also give you unforseen results.  So I like to err on the side of strict typing by default with the option of loose type checking as a feature if I need it.

fprochazka: using strict comparision is best practice

PhpAmqpLib/Message/AMQPMessage.php
postalservice14: @videlalvaro What parent class?  `GenericContent` doesn't seem to have `setBody`

postalservice14:
I just pushed some more changes, most are from https://github.com/videlalvaro/php-amqplib/pull/200 
Just trying to clean up some old pull requests.  Thanks again @eexit

eexit:
@postalservice14 @videlalvaro I'm rebasing my PR, wait for it please.

videlalvaro:
@postalservice14 #201 has been merged

postalservice14:
@videlalvaro this should be good to go now.  I snuck in a fix to travis and scrutinizer with PHP 7 that I introduced.  Apparently xdebug is not installed on PHP 7 and so there is no code-coverage reports to send to scrutinizer.

PhpAmqpLib/Channel/AMQPChannel.php
```diff
@@ -193,9 +192,9 @@ protected function channel_close($args)
      * Confirm a channel close
      * Alias of AMQPChannel::do_close()
      *
-     * @param AMQPReader $args
+     * @param AMQPReader $reader
      */
-    protected function channel_close_ok($args)
+    protected function channel_close_ok(AMQPReader $reader)
```
videlalvaro: just to be sure, wouldn't this introduce some breaking changes?

postalservice14: How so?   The only way this would break, is if the argument wasn't of type `AMQPReader`

.scrutinizer.yml
```diff
@@ -1,6 +1,6 @@
 filter:
     paths: [PhpAmqpLib/*]
-    excluded_paths: [tests/*, demo/*, benchmark/*, spec/*]
+    excluded_paths: [tests/*, demo/*, benchmark/*, spec/*, PhpAmqpLib/Helper/Protocol/*]
```
postalservice14: This is because these files are generated and we have little control over how they are formatted.

PhpAmqpLib/Channel/AMQPChannel.php
postalservice14: If you're worried about it, I can remove the type-hint and just leave the PHPDoc block

eexit: See https://github.com/videlalvaro/php-amqplib/pull/200#issuecomment-64799831

postalservice14:
@videlalvaro I see, I've removed the type-hints.  Sounds like something that might want to be introduced in 3.*

videlalvaro:
Sorry for the delay. I'll take a look later

PR #325, opened 2015-10-06
Support PHP 7.0

postalservice14:
It looks like, in it's current state, the library supports and passes PHP 7.  Since it will be released soon, do we want to remove PHP 7 from the allowed "failures section" of TravisCI?


videlalvaro:
I'd assume PHP 7.0 tests are passing, right?

postalservice14:
Yep.  Scrutinizer failed, I think, because it was working on another PR that got merged before it finished.

PR #326, opened 2015-10-06
Use new TravisCI infrastructure

postalservice14:
In order to do this, I had to have travis use the default vhost and username/password.  Hoping that's not a problem.


videlalvaro:
thanks

PR #328, opened 2015-10-15
just removing an unnecessary else

ThaDafinser:


videlalvaro:
Can you expand on why the else is unnecessary and what are you trying to solve? The whitespace changes doesn't seem to be related to this PR either, so it gets quite confusing to understand what to review

PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -559,26 +559,25 @@ protected function wait_channel($channel_id, $timeout = 0)
                 }
                 continue;
 
-            } else {
```
ThaDafinser: @videlalvaro i mean this else. Because when you use `return` the code is stopped at that point, so no else is needed

videlalvaro: I think this is gone from master branch, am I correct?

ThaDafinser: Hui it's this part: https://github.com/videlalvaro/php-amqplib/blob/master/PhpAmqpLib/Connection/AbstractConnection.php#L560-L562

In the above `if` you use `continue` so the following `else is not needed

videlalvaro: AFAIK the `continue` will make the `while` loop above to jump to the next iteration, so this is not related to the `else` below

ThaDafinser: Yes it jumps to the next iteration -> so it continues with the next element on the top of the `while` .

That's why you need no else at all.

videlalvaro: No, the `else` is there in case `if ($frame_channel === 0 && $frame_type === 8)` doesn't match. The `else` and the `continue` are totally unrelated

videlalvaro: What could be removed is the `continue`

ThaDafinser: @videlalvaro you can remove the `else` or the `continue` ...in my PR i wanted to remove the `else` because you remove one level.

See this example - it's exactly the same behaviour
```php
while(true){
  if($i === 2){
    // do this
  } else{
    // do something elese
  }
}

// this do the same
while(true){
  if($i === 2){
    // do this
    continue;
  }

  // do something elese
}
```

videlalvaro: I think removing the `continue` is clearing in the intention of what's going on in the code.

PR #336, opened 2015-11-02
Remove hhvm-nightly and fix PHP 7 causing scrutinizer errors

postalservice14:
PHP 7 does not publish code-coverage.  I think that's what is breaking the scrutinizer builds.

Also, hhvm-nightly is no longer supported on Travis.


postalservice14:
@videlalvaro This should fix the scrutinizer build issues.

Issue #322, opened 2015-10-01
Memory exhausted in StreamIO.php


ruudk:
I'm using version 2.6 of php-amqplib with the latest version of RabbitMqBundle and for weeks I run into a fatal memory exhausted error. It is trying to allocate 1.34 gigabyte in StreamIO.php on line 194.

I did a lot of debugging and research to find the issue causing this.

I use RabbitMQ 3.5.4 with the following RabbitMqBundle settings:
- keepalive: false
- heartbeat: 30
- read_write_timeout: 60
- Consumer argument: `--memory-limit=350`

My PHP has a memory limit of 500MB.

After every received message I do a memory cleanup and display the memory in use:

```
Oct 01 09:38:40 aws-pdf-web1 eventbus-2.log:  Auto cleanup, 59.5 mb in use.
Oct 01 09:53:21 aws-pdf-web1 eventbus-2.log:  Fatal error: Allowed memory size of 524288000 bytes exhausted (tried to allocate 1342177291 bytes) in /home/ticketswap/app/releases/20150930160705/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php on line 194
```

As you can see, after receiving the last message it did nothing for 15 minutes and then failed.

Any thoughts what might cause this issue?

videlalvaro:
maybe this is related to this? https://github.com/videlalvaro/php-amqplib/pull/319

ruudk:
Just deployed that fix to production. The error normally occurs at night when there is less activity, so fingers crossed :)

videlalvaro:
just merged it BTW

bmcgavin:
Hi, hope this isn't considered derailing.

I've had the same error (specifically the number of bytes, 'trying to allocate 1342177291 bytes'), and found this link helpful in explaining what's going on : 

http://john.eckersberg.com/debugging-rabbitmq-frame_too_large-error.html

I think the protocol handling gets into an incorrect state if there's a networking problem. I've simulated this with a consumer (using AMQPStreamConnection) and iptables.

* Consumer starts with a heartbeat of 2 seconds
* Drop packets from the consumer to RMQ : iptables -A OUTPUT -p tcp --dport 5672 -j DROP
* Sleep 10 seconds
* Drop that rule to re-enable comms : iptables -D OUTPUT 1

[...]

PR #338, opened 2015-11-19
Throw an exception if would allocate too much memory for a frame

bmcgavin:
As in my comment on #322, the broker can send a message that causes the client to allocate memory to read a 1.25GB frame, usually causing an out of memory error.

This tries to mitigate this by throwing an exception if more than 8MB of memory would be allocated, allowing the client to better negotiate a reconnection.

### DIFF
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -51,6 +51,9 @@ class StreamIO extends AbstractIO
     /** @var bool */
     private $canDispatchPcntlSignal;
 
+    /** @var int */
+    private $amqpHeaderAsInt;
+
     /**
      * @param string $host
      * @param int $port
@@ -59,6 +62,7 @@ class StreamIO extends AbstractIO
      * @param null $context
      * @param bool $keepalive
      * @param int $heartbeat
+     * @param int $amqpHeaderAsInt
      */
     public function __construct(
         $host,
@@ -67,7 +71,8 @@ class StreamIO extends AbstractIO
         $read_write_timeout,
         $context = null,
         $keepalive = false,
-        $heartbeat = 0
+        $heartbeat = 0,
+        $amqpHeaderAsInt = 1342177291
     ) {
         $this->protocol = 'tcp';
         $this->host = $host;
@@ -79,6 +84,11 @@ class StreamIO extends AbstractIO
         $this->heartbeat = $heartbeat;
         $this->canSelectNull = true;
         $this->canDispatchPcntlSignal = $this->isPcntlSignalEnabled();
+
+        //Used to prevent an OOM error when the client re-sends the AMQP header
+        // to an already-connected broker
+        //See http://john.eckersberg.com/debugging-rabbitmq-frame_too_large-error.html
+        $this->amqpHeaderAsInt = $amqpHeaderAsInt;
 
         if (is_null($this->context)) {
             $this->context = stream_context_create();
@@ -198,8 +208,8 @@ class StreamIO extends AbstractIO
                 throw new AMQPRuntimeException('Broken pipe or closed connection');
             }
 
-            if (($len - $read) > 8 * (1024 * 1024)) {
-                throw new AMQPRuntimeException('Trying to allocate 8M, aborting');
+            if (($len - $read) >= $this->amqpHeaderAsInt) {
+                throw new AMQPRuntimeException('Frame too large');
             }
 
             set_error_handler(array($this, 'error_handler'));
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -198,6 +198,10 @@ class StreamIO extends AbstractIO
                 throw new AMQPRuntimeException('Broken pipe or closed connection');
             }
 
+            if (($len - $read) > 8 * (1024 * 1024)) {
+                throw new AMQPRuntimeException('Trying to allocate 8M, aborting');
+            }
+
             set_error_handler(array($this, 'error_handler'));
             $buffer = fread($this->sock, ($len - $read));
             restore_error_handler();


postalservice14:
Can you help me understand the significance of the 8MB?  I guess I'm not understanding why allocating more than 8MB is a bad thing.  Thanks for the PR!

bmcgavin:
Hi, thanks for the peer review! I chose 8mb as an arbitrary number to be honest, but one could easily say 1gb and avoid the problem.

I assumed that a message would not be bigger than 8mb, but maybe that is a fallacy

postalservice14:
Yeah, I'm not sure we can assume that unfortunately :-/  I wonder if we can check to see if the message is greater than the allocated memory for PHP or something?

bmcgavin:
Do you think it would be okay to say if the message was 1gb then reject it?

postalservice14:
I think if we did anything we need to make it configurable.  Have it as a `max_message_size` that defaults to 500MB or something.  500MB seems reasonable since I think the PHP default memory limit is 512MB.

bmcgavin:
Added another commit to make the trigger point configurable, and it defaults to the number I saw in the initial problem.
I'll test that this fixes the problem on Monday.

PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -59,6 +62,7 @@ class StreamIO extends AbstractIO
      * @param null $context
      * @param bool $keepalive
      * @param int $heartbeat
+     * @param int $amqpHeaderAsInt
```
postalservice14: I think it would be better for optional parameters to not be on the constructor and use a setter instead.

Also would it make sense to name it something like `$maxFrameSize`?

postalservice14:
What are your thoughts on something like below?:

```diff
Index: PhpAmqpLib/Wire/IO/StreamIO.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- PhpAmqpLib/Wire/IO/StreamIO.php	(revision a0b99e0b5d728a33c419617d18625ae27c5721f3)
+++ PhpAmqpLib/Wire/IO/StreamIO.php	(revision )
@@ -52,6 +52,15 @@
     private $canDispatchPcntlSignal;
 
     /**
+     * Maximum allowable size of individual frames, in bytes.
+     *
+     * Defaults to 1.5GB.
+     *
+     * @var int
+     */
+    private $maxFrameSize = 1.5e+9;
+
+    /**
      * @param string $host
      * @param int $port
      * @param int $connection_timeout
@@ -198,6 +207,10 @@
                 throw new AMQPRuntimeException('Broken pipe or closed connection');
             }
 
+            if (($len - $read) >= $this->maxFrameSize) {
+                throw new AMQPRuntimeException('Frame too large');
+            }
+
             set_error_handler(array($this, 'error_handler'));
             $buffer = fread($this->sock, ($len - $read));
             restore_error_handler();
@@ -436,5 +449,15 @@
         $this->heartbeat = 0;
 
         return $this;
+    }
+
+    /**
+     * Set the maximum size per frame in bytes.
+     *
+     * @param int $maxFrameSize
+     */
+    public function setMaxFrameSize($maxFrameSize)
+    {
+        $this->maxFrameSize = $maxFrameSize;
     }
 }
```

bmcgavin:
Yes, that would do it.

michaelklishin:
Is it just me or the idea of `max frame size` is misinterpreted in this thread? It is not the max message size but rather the size of a single message payload frame. It's expected that clients will frame larger payloads. Pretty much every client uses 128 Kb as max frame size (this value is not particularly scientific; however, most messages are a few kilobytes in size, so it is good enough in practice).

I'm yet to see a case where avoiding body framing is actually necessary, in any RabbitMQ client.

Also: RabbitMQ forces a 2 GB limit on message size (framing does not change this).

postalservice14:
@michaelklishin thanks for jumping in.  I don't think that's being misinterpreted, but I'm not sure I fully understand what causes such a large frame size?  You mention pretty much every client uses 128kb.  That sounds good to me.  So do you you think just limiting the read amount to less than 128kb would fix this problem?

@bmcgavin can you try the code below with your issue and see if this fixes it?  If it does, can you update your PR to the code below?:

```diff
Index: PhpAmqpLib/Wire/IO/StreamIO.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- PhpAmqpLib/Wire/IO/StreamIO.php	(revision a0b99e0b5d728a33c419617d18625ae27c5721f3)
+++ PhpAmqpLib/Wire/IO/StreamIO.php	(revision )
@@ -9,6 +9,11 @@
 
 class StreamIO extends AbstractIO
 {
+    /**
+     * Maximum frame size in bytes.
+     */
+    const MAX_FRAME_SIZE = 128000;
+
     /** @var string */
     protected $protocol;
 
@@ -198,8 +203,10 @@
                 throw new AMQPRuntimeException('Broken pipe or closed connection');
             }
 
+            $frameSize = ($len - $read) < static::MAX_FRAME_SIZE ? ($len - $read) : static::MAX_FRAME_SIZE;
+
             set_error_handler(array($this, 'error_handler'));
-            $buffer = fread($this->sock, ($len - $read));
+            $buffer = fread($this->sock, $frameSize);
             restore_error_handler();
 
             if ($buffer === false) {
```

bmcgavin:
Hi @postalservice14, this won't fix it, because the protocol error causes ($len - $read) to be ~1342177291 bytes, which is larger than the maximum frame size.

To re-iterate, this is not a problem with the handling of valid messages. This is a problem where (I think) the client sends a message to the broker that causes the broker to send a protocol header to the client, which misinterprets the ASCII bytes 'P009' as the frame length, causing the overflow.

I'll have to re-investigate the problem, as I can't really recall the details from November. It may be avoidable with additional checks in the protocol handling code rather than in the message reading logic.

Thanks for the attention so far, I'm glad the library is still being maintained.

michaelklishin:
@postalservice14 if this client can reconstruct body frames into a single payload correctly (before passing it on to consumers), then yes. Frame length is indicated in its "prefix", so there is no need to always allocate `frame_max` bytes before doing a read.

postalservice14:
@michaelklishin @bmcgavin I'm torn on what to do about this problem.

Can this issue be solved just by setting the `frame_max` setting in RabbitMQ to the default (Looks like `131072` bytes)?  I'm still a little confused how the client is getting such a large frame, unless the `frame_max` is not the default and is set to a very large number.

> This is a problem where (I think) the client sends a message to the broker that causes the broker to send a protocol header to the client, which misinterprets the ASCII bytes 'P009' as the frame length, causing the overflow.

So you think the problem is on the client sending the message?  So it's not a php-amqplib problem?

Thanks guys.

michaelklishin:
@postalservice14 I don't understand the quoted part. RabbitMQ does not send protocol header to clients; clients send it right after opening TCP connection.

128 KB is a much more sensible `frame_max` value. There can be other issues and inefficiencies around framing implementation of this client, though. And no, you can't blame other clients for how this one does framing.

postalservice14:
Yeah, I don't really understand the quoted part either.  That's why I asked the questions after it.

I think what would help avoid confusion is if @bmcgavin was able to write a test or a way to reproduce it.

bmcgavin:
Hi @postalservice14, @michaelklishin - there's a way to reproduce on the linked issue in my initial comment, and I've been able to re-reproduce this today.

* Start a consumer with a 2s heartbeat, 3s connection timeout
* Do something with iptables to prevent the heartbeats completing (today I dropped incoming traffic on 5672 on the RMQ broker)
* Wait until the heartbeat timeout triggers a reconnect attempt (caused by AMQPChannel->wait() throwing an exception)
* Wait until the reconnect attempt fails (caused by AMQPStreamConnection->reconnect() throwing an exception once the connect_timeout is exceeded)
* Remove the iptables rule - the consumer will get partway through the reconnection but the broker will send the AMQP header causing the large frame allocation.

You can see the consumer script I used [here](https://gist.github.com/bmcgavin/19d66dcf55b932fef7e9), the tcpdump output [here](https://gist.github.com/bmcgavin/43c7ebe7bb591b1d6219) and the logs with OOM error [here](https://gist.github.com/bmcgavin/789431ed75fca81265ee).

It may be the case that my consumer code is incorrect, but I would still rather catch the large allocation in the library as there's no real way to mitigate against it in the client code.

In fact given the stacktrace in the gist, I can see the problem. If the broker does send the header (when an established connection is broken and then re-established), but the library is still in the wait() call, then this code in Connection\AbstractConnection will cause the problem :

```
$frame_type = $this->wait_frame_reader->read_octet();
$channel = $this->wait_frame_reader->read_short();
$size = $this->wait_frame_reader->read_long();

// payload + ch
$this->wait_frame_reader->reuse($this->input->read(AMQPReader::OCTET + (int) $size));

$payload = $this->wait_frame_reader->read($size);
```

`$frame_type` will be 'A', `$channel` will be 'MQ' and `$size` will be 0x50000009, or 1342177289 bytes. The next `$input->read()` call will overflow.

As 'A' is an invalid frame type, this problem could be fixed here.

bmcgavin:
I'll close this one as the fix is messy - #372 is the result of more thorough research.

PR #339, opened 2015-11-19
fix heartbeat

woffs:
when the network connection is broken, the select never returns and
never recognizes the missing heartbeat. changing the select (null) to
select (heartbeat+1s) hotfixes this.

Tested with php-5.4.45 on Debian Wheezy and an iptables rule to block AMQP traffic. Tested with php-amqplib-v2.6.0, but rebased to master.


woffs:
Seems that this is without effect in HEAD. Heartbeat is still broken.

PR #340, opened 2015-11-20
StreamIO: fix "bad write retry" in SSL mode with nonblocking I/O

fuhry:
When a large frame is written to the wire over an SSL connection, PHP's OpenSSL extension behaves erroneously upon write failures in non-blocking mode. The fwrite call will return 0. However, if the variable containing the buffer is modified or reassigned, PHP will pass a different pointer to SSL_write(), causing the "bad write retry" error.

This behavior is also produced when a large frame (>= 49,152 bytes) is written directly to the wire with fwrite().

This patch resolves the issue by changing the following behavior:

  - Limits fwrite() to 8,192 bytes at a time
  - Does not reassign $data if 0 bytes are written

No API breakage. Unit tests are passing.

Datto is using php-amqplib internally and so far we have seen reliability greatly improved with this patch.

### DIFF
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -255,8 +255,15 @@ class StreamIO extends AbstractIO
                 throw new AMQPRuntimeException('Broken pipe or closed connection');
             }
 
-            set_error_handler(array($this, 'error_handler'));
-            $buffer = fwrite($this->sock, $data, 8192);
+			set_error_handler(array($this, 'error_handler'));
+			// OpenSSL's C library function SSL_write() can balk on buffers > 8192
+			// bytes in length, so we're limiting the write size here. On both TLS
+			// and plaintext connections, the write loop will continue until the
+			// buffer has been fully written.
+			// This behavior has been observed in OpenSSL dating back to at least
+			// September 2002:
+			// http://comments.gmane.org/gmane.comp.encryption.openssl.user/4361
+			$buffer = fwrite($this->sock, $data, 8192);
             restore_error_handler();
 
             if ($buffer === false) {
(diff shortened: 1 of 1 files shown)


PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -256,7 +256,7 @@ public function write($data)
             }
 
             set_error_handler(array($this, 'error_handler'));
-            $buffer = fwrite($this->sock, $data);
+            $buffer = fwrite($this->sock, $data, 8192);
```
postalservice14: What's the significance of 8192 being the number chosen?  Can we make this a constant or something to help describe what this number is?  Just a thought.

postalservice14:
+1 We really need this, thanks for the PR.

PhpAmqpLib/Wire/IO/StreamIO.php
fuhry: Thanks for the feedback. OpenSSL's SSL_write() function has been known to misbehave for quite some time now when passed a buffer larger than 8,192 bytes in length (see [this mailing list post](http://comments.gmane.org/gmane.comp.encryption.openssl.user/4361) from September, 2002). I'll add a comment to the source explaining this.

fuhry:
@postalservice14 @videlalvaro All set on suggested changes. Is this ready to be merged?

videlalvaro:
thanks

finalgamer:
This patch seems to drastically decrease the speed of the connection.

I am sending 10MB to my local RabbitMQ Server witch takes me 6.5s.
If i remove the limit on fwrite() the 10MB take 0.1s.

It would be nice if we can get some sort of config to adjust if we want to do the limiting.
Or only do it for SSL connections.

PR #341, opened 2015-11-26
AMQPReader::wait() throw IOWaitException on stream_select failure

holycheater:
A follow-up on this:
https://github.com/videlalvaro/php-amqplib/pull/264
And this:
https://github.com/videlalvaro/php-amqplib/pull/224

amqplib now processes stream_select warning. But it returns false. And AMQPReader throws AMQPRuntimeException.
My implementation did re-run stream_select with getting whatever it returns.

I need to catch an exception when stream_select fails so I can call upon channel wait again, but AMQPRuntimeException is a very general one and filtering exception by its message text is not a good practice.
I propose to throw different exception so you can catch stream_select failure and do something about it. It extends from AMQPRuntimeException so it doesn't break library behavior.

### DIFF
--- a/PhpAmqpLib/Exception/AMQPIOWaitException.php
+++ b/PhpAmqpLib/Exception/AMQPIOWaitException.php
@@ -1,0 +1,6 @@
+<?php
+namespace PhpAmqpLib\Exception;
+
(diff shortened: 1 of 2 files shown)


postalservice14:
Thanks!

PR #345, opened 2015-12-11
Float timeout for SteamIO and SocketIO

todoshcenko:


ob-ivan:
:+1:

videlalvaro:
How does it affect BC?

todoshcenko:
There's a problem with backward compat, indeed. If a client sets a 0.01 timeout (by any reason) then the timeout gets disabled in the current master (because of implicit rounding). With this change these timeouts will be re-enabled, which may be unexpected for the client.

todoshcenko:
Backward compatibility implemented.

This is also to draw attention of those clients who use float value timeouts that they possibly didn't get what they wanted.

todoshcenko:
Do you нave any comments or suggestions on this solution?

postalservice14:
Two things:

1. Just out of curiosity, why do you need float precision for timeouts?
2. Why do you have the `AMQP_FLOAT_TIMEOUT_FOR_SOCKET` constant?  Can't you just detect if it's a float or not using `is_float`?

ob-ivan:
@postalservice14 Consider a client asking a service for some data which is not critical for the client's job and may be omitted if it takes too much time for the service to respond. And "too much" means 100ms, for example. In a highloaded project there's a bunch of such clients, say thousands of them firing requests simultaneously. If each client were to open a connection for the whole second, it wouldn't take long for server to run out of system resources --- for useless 900ms each waiting for something that doesn't matter anymore. So you'd prefer to cut down connection timeout to the said 100ms and spare a lot of server resources.

postalservice14:
Makes perfect sense thanks.

todoshcenko:
@postalservice14 Using `is_float` broke BC. 
If a client sets a 0.01 timeout (by any reason) then the timeout gets disabled in the current master (because of implicit rounding). With this change these timeouts will be re-enabled, which may be unexpected for the client. In current solution this clients gets `E_USER_DEPRECATED` instead of unexpected behaviour.

postalservice14:
@todoshcenko I'm torn on this one a little bit.  Community, please let me know your thoughts.

On one hand, I don't think it's a BC break, because `$timeout` everywhere is specified as an int.  If someone were passing in a float already (the `0.01` scenario), it's technically in invalid $timeout that we just weren't correctly catching.
On the other hand, PHP does allow for this....however right or wrong people thing it is...

I'm leaning toward the former argument, and if someone's code is passing in a `0.01` expecting it to disable the timeout, they are already misusing the $timeout argument.  Also, we can put it in the changelog as a "possible but unlikely BC break"

I'm just not loving the configuration by constant `AMQP_FLOAT_TIMEOUT_FOR_SOCKET` approach too much.

Community, what are your thoughts?

todoshcenko:
@postalservice14 Personally I don't like it myself. We can rollback to the first solution if the "possible but unlikely BC break" is OK with the project's policies.

postalservice14:
@nubeiro What do you think?  I'm still leaning towards the solution that has the "possible but unlikely BC break".

therosco:
Is any progress there?

nubeiro:
I agree that passing 0.01 as value for timeout should be considered misusage, it is a possible BC break, but as you all have said, it is unlikely. 

This adds an enhancement to fine tune timeouts, I'd say we go for it.

therosco:
I really appreciate your decision.

Should we expect a release containing this enhancement in a couple of days?

todoshcenko:
Finaly:

- the code is made simple again (doesn't have ugly constant AMQP_FLOAT_TIMEOUT_FOR_SOCKET and warnings).
- the code has possible but unlikely BC break.

Can we merge it now?

nubeiro:
I'll dedicate part of the weekend to this and other open issues

PR #346, opened 2015-12-11
handle connection interruption in v2.6.0

woffs:
In a consumer, when the network connection is broken, select() never returns and
never recognizes the missing heartbeat. changing the select (null) to
select (heartbeat+1s) hotfixes this.

In a publisher, broken network connection is not recognized. Implementing write timeout fixes this.


videlalvaro:
There are conflicts on this branch

woffs:
see  #348

PR #347, opened 2015-12-14
FIXED AbstractCollection value de/encoding on PHP7

bhelm:
the old decodeCollection function returned an array with uninitialized values. Changing the foreach loop to not use references solved this on PHP7.


videlalvaro:
thanks

PR #358, opened 2016-01-27
Fix issue with declaring an exchange with a closed connection

### LINKED ISSUE #337: Fatal error: Call to a member function send_channel_method_frame() on null
Test code:

<?
// std connection routines

// check exchange exists
try {
	$this->channel->exchange_declare('tst.exchange2', 'topic', true);
	// exists - make something
} catch (\Exception $e) {
	// not exist
}
$this->channel->exchange_declare('tst.exchange2', 'topic', false, true, false);
?>

And on the last line we get the fatal error: 
Fatal error: Call to a member function send_channel_method_frame() on null

postalservice14:
Fixes #337

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_script:
   - travis_retry composer update ${COMPOSER_FLAGS} --no-interaction --prefer-source
 
 script:
-  - vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover
+  - vendor/bin/phpunit -d zend.enable_gc=0 --coverage-text --coverage-clover=coverage.clover
 
 after_script:
   - if [[ $TRAVIS_PHP_VERSION != 'hhvm' ]]; then php vendor/bin/ocular code-coverage:upload --format=php-clover coverage.clover; fi
--- a/PhpAmqpLib/Channel/AbstractChannel.php
+++ b/PhpAmqpLib/Channel/AbstractChannel.php
@@ -223,6 +223,10 @@ abstract class AbstractChannel
      */
     protected function send_method_frame($method_sig, $args = '')
     {
+        if ($this->connection === null) {
+            throw new AMQPRuntimeException('Channel connection is closed.');
+        }
+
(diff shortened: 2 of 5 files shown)


videlalvaro:
I think this error can happen for more situations that just exchange declare, what do you think?

postalservice14:
Yeah, it definitely can, I was just pushing out a quick fix.  I can add more scenarios...

videlalvaro:
I think a key to maintaining the library would be to find more generic solutions based on these quick fixes, so in the long term the library doesn't become a big patch of various things, which sadly it had become somehow

PhpAmqpLib/Channel/AMQPChannel.php
```diff
@@ -342,6 +341,10 @@ public function exchange_declare(
         $arguments = null,
         $ticket = null
     ) {
+        if (!$this->is_open) {
+            throw new AMQPRuntimeException('Channel connection is closed.');
```
michaelklishin: What does "channel connection" mean? What specifically is closed, a channel or a connection?

videlalvaro: I'd guess it means the connection to which this channel belong, has been closed

michaelklishin: Java, .NET, and Ruby (Bunny) clients says something like "an operation was attempted on a closed channel."

I guess in PHP connection and channel lifecycles are often inseparable, I recommend that the message focuses on channel being closed. It can be due to a channel-level protocol exception or connection closure or TCP connection loss: the end result for the caller is the same.

postalservice14:
@videlalvaro I agree.  I've updated the PR to check on the `send_method_frame` function, so we can catch it for multiple scenarios.

PhpAmqpLib/Channel/AMQPChannel.php
postalservice14: The end result is the same, but the message I currently have, IMO, is more accurate.  It gives the exact reason why the operation could not be performed, which helps with debugging and such.

Issue #248, opened 2015-02-24
AMPConnection just sits there


BardiaAfshin:
I think this is the right place for it, in my code when I use AMPQConnection to connect to a local rabbitmq server, all things are swell. However, when I bring this up on AWS Services ec2 instance running rabbit mq, it's not working.

``` php
try {
    $connection = new AMQPConnection('myinstance-at-amazon', 15672, 'user', 'password');
} catch (Exception $e) {
    echo 'Caught exception: ', $e->getMessage(), "\n";
}
```

When I run this at command line, from my echo statements I can see the the connection is attempted, however it just hangs out there.

I thought it's a firewall of sorts configuration, something blocking the outbound connection, but again, that's not the case. I took a look at netstat on both the client machine attempting to make the connection, and the instance running rabbit mq, there's activity when I fire up the attempted connection, but again NO TIMEOUT! AND NO EXCEPTION!!!

I can see from the construct of AMPQConnection that there is a value of '3' given for timeout, but it's just not timing out.

Help? Much appreciated.

Update:
As a process of elimination, I was able to connect just fine using python pika library, here's the code, from AMPQ vanilla examples

``` python
#!/usr/bin/env python
import pika


connection = pika.BlockingConnection(pika.ConnectionParameters(
        host='my-instance.amazonaws.com',
        credentials=pika.credentials.PlainCredentials('asdf', 'secret')))
channel = connection.channel()

channel.queue_declare(queue='hello')

channel.basic_publish(exchange='',
                      routing_key='hello',
                      body='Hello World!')
print " [x] Sent 'Hello World!'"
connection.close()
```

but, the receiver pika version isn't working, hmmm

``` python
#!/usr/bin/env python
import pika
import time

connection = pika.BlockingConnection(pika.ConnectionParameters(
        host='my-instance.amazonaws.com',
        credentials=pika.credentials.PlainCredentials('asdf', 'asdf')))
channel = connection.channel()

channel.queue_declare(queue='task_queue', durable=True)
print ' [*] Waiting for messages. To exit press CTRL+C'

def callback(ch, method, properties, body):
    print " [x] Received %r" % (body,)
    time.sleep( body.count('.') )
    print " [x] Done"
    ch.basic_ack(delivery_tag = method.delivery_tag)

channel.basic_qos(prefetch_count=1)
channel.basic_consume(callback,
                      queue='task_queue')

channel.start_consuming()
```

michaelklishin:
[...]

PR #360, opened 2016-02-03
Fix issue where connection timeout hangs the script

postalservice14:
Fixes: #248


postalservice14:
Need to find a different way to do this.  This breaks when waiting for new frame.

PR #366, opened 2016-02-11
Add Contributor Code of Conduct

postalservice14:
Thoughts or concerns about adding this CoC to the project?  Please avoid the "drama"...  :-)

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,49 @@
+# Contributor Code of Conduct
+
+As contributors and maintainers of this project, and in the interest of
+fostering an open and welcoming community, we pledge to respect all people who
+contribute through reporting issues, posting feature requests, updating
+documentation, submitting pull requests or patches, and other activities.
+
+We are committed to making participation in this project a harassment-free
+experience for everyone, regardless of level of experience, gender, gender
+identity and expression, sexual orientation, disability, personal appearance,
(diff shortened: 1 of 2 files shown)


videlalvaro:
What's the suggested way of reporting violations? It has to be private.

postalservice14:
Yeah, I wasn't sure how to handle that... :-/  Any ideas?  I didn't want to put just my email address, that didn't seem right...

videlalvaro:
Yeah. Let's ask github support to add a "report" button of sorts. If we all three ask they might pay attention

nubeiro:
I agree, violation reports must be private, so that we can try to solve issues between implied parties.

nubeiro:
If requesting a report button does not work, we might try to use what @postalservice14 suggested, which is letting the reporter try to find out how to contact us... our e-mail addresses are in composer.json file. If that is not enough, we could consider adding Twitter accounts in readme file or even mentioning we use Gitter as IM tool.

postalservice14:
I sent a support request to Github as well....

For the time being, do we want to leave it as is?

nubeiro:
I got reply today that they would " pass your suggestion to the team to consider" but no promises made on actually getting something done. 
I'd leave it as it is, if we ever get a "private report" button then we can update CoC

postalservice14:
Sounds good to me.

videlalvaro:
I've got the same reply from them

PR #369, opened 2016-02-12
Cleanup tests and make them run much faster

postalservice14:
Build go from 7-10 minutes to ~1.5 minutes.

Basically among some basic cleanup, I removed the loops in the WireTest class and replace them with the testing the edges (minimums and max's plus or minus 1 value).  That should give us pretty good test coverage without running through ~532000 assertions.

So to help those reviewing.  All changes are formatting and variable name changes, except for `WireTest`


PhpAmqpLib/Wire/AMQPWriter.php
```diff
@@ -180,28 +180,32 @@ public function write_signed_octet($n)
     /**
      * Write an integer as an unsigned 16-bit value
      *
-     * @param $n
+     * @param int $shortValue
      * @return $this
      * @throws \PhpAmqpLib\Exception\AMQPInvalidArgumentException
      */
-    public function write_short($n)
+    public function write_short($shortValue)
```
videlalvaro: Something we always did at RabbitMQ was to have PRs be about one particular change. The PR mentions tests clean up but at the same time these variables are changed. I'm not sure this is the right place for these changes. Also a different PR for this would make tracing changes in the repo history easier

postalservice14: Yeah, good or bad...it's a habit that I see a variable name that doesn't make sense and I change it. :-/  I can split it up....

videlalvaro: I think the change is good, but we should have the discipline of splitting PRs. Same will happen when you receive contributions later and they will have similar problems. So we have to try at least to set an example (not that I've never did this BTW)

postalservice14: Ha, no worries.  Very valid and good point.  Thank you...

The change is made and rebased btw.

michaelklishin:
@postalservice14 you've been on a roll with pull requests and issues in this project lately. Thank you on behalf of the RabbitMQ team :+1:

postalservice14:
Thanks @michaelklishin .  Doing my best... :-)

PR #372, opened 2016-02-17
Get the array of valid frame types from the protocol constants class

bmcgavin:
A neater and more concise fix for #338.

When a frame_type is read from a message, get the list of valid frame types for the protocol version in use, and validate against them, throwing an AMQPRuntimeException if invalid.


postalservice14:
This fixes #383?  Awesome!  This is much more concise.  I like it!

PR #375, opened 2016-02-19
Add Missing DocBlocks and Types

rcatlin:
* Added missing doc blocks
* Added missing parameter/return types in doc blocks.

Most integer parameter types were hinted using `int` rather than `integer`, and booleans as `bool` rather than `boolean`. I changed any `boolean` types to `bool` and `integer` types to `int` to stay consistent.

### DIFF
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -644,7 +644,7 @@ class AMQPChannel extends AbstractChannel
      * Confirms a queue definition
      *
      * @param AMQPReader $reader
-     * @return string[]
+     * @return array
      */
     protected function queue_declare_ok($reader)
     {
--- a/PhpAmqpLib/Channel/AbstractChannel.php
+++ b/PhpAmqpLib/Channel/AbstractChannel.php
@@ -170,7 +170,7 @@ abstract class AbstractChannel
     /**
      * @param string $method_sig
      * @param string $args
-     * @param $content
+     * @param mixed $content
      * @return mixed
      * @throws \PhpAmqpLib\Exception\AMQPRuntimeException
      */
@@ -218,7 +218,7 @@ abstract class AbstractChannel
     }
 
     /**
-     * @param $method_sig
+     * @param string $method_sig
      * @param string $args
      */
     protected function send_method_frame($method_sig, $args = '')
@@ -347,6 +347,10 @@ abstract class AbstractChannel
         }
     }
 
+    /**
+     * @param string $allowed_methods
+     * @return array
+     */
     protected function process_deferred_methods($allowed_methods)
     {
         $dispatch = false;
@@ -368,6 +372,10 @@ abstract class AbstractChannel
         return array('dispatch' => $dispatch, 'queued_method' => $queued_method);
     }
 
+    /**
+     * @param array $queued_method
+     * @return mixed
+     */
     protected function dispatch_deferred_method($queued_method)
     {
         $this->debug->debug_method_signature('Executing queued method: %s', $queued_method[0]);
@@ -376,7 +384,7 @@ abstract class AbstractChannel
     }
 
     /**
-     * @param $frame_type
+     * @param int $frame_type
      * @throws \PhpAmqpLib\Exception\AMQPRuntimeException
      */
     protected function validate_method_frame($frame_type)
@@ -384,11 +392,19 @@ abstract class AbstractChannel
         $this->validate_frame($frame_type, 1, 'AMQP method');
     }
 
+    /**
+     * @param int $frame_type
+     * @throws \PhpAmqpLib\Exception\AMQPRuntimeException
+     */
     protected function validate_header_frame($frame_type)
     {
         $this->validate_frame($frame_type, 2, 'AMQP Content header');
     }
 
+    /**
+     * @param int $frame_type
+     * @throws \PhpAmqpLib\Exception\AMQPRuntimeException
+     */
     protected function validate_body_frame($frame_type)
     {
         $this->validate_frame($frame_type, 3, 'AMQP Content body');
@@ -413,7 +429,7 @@ abstract class AbstractChannel
     }
 
     /**
-     * @param $payload
+     * @param string $payload
      * @throws \PhpAmqpLib\Exception\AMQPOutOfBoundsException
      */
     protected function validate_frame_payload($payload)
@@ -423,6 +439,10 @@ abstract class AbstractChannel
         }
     }
 
+    /**
+     * @param string $payload
+     * @return string
+     */
     protected function build_method_signature($payload)
     {
         $method_sig_array = unpack('n2', mb_substr($payload, 0, 4, 'ASCII'));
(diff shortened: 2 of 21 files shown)


rcatlin:
@postalservice14 Review? I combed through and added any missing doc blocks and types I could find. This should be all of them.

PhpAmqpLib/Channel/AbstractChannel.php
```diff
@@ -170,7 +170,7 @@ public function getMethodQueue()
     /**
      * @param string $method_sig
      * @param string $args
-     * @param $content
+     * @param mixed $content
```
postalservice14: I believe `$content` is always of type `AMQPMessage`.  In fact, we could probably rename the variable `$amqpMessage`

PhpAmqpLib/Channel/AbstractChannel.php
```diff
@@ -218,7 +218,7 @@ public function next_frame($timeout = 0)
     }
 
     /**
-     * @param $method_sig
+     * @param string $method_sig
```
postalservice14: `$method_sig` is actually an array.

PhpAmqpLib/Channel/AbstractChannel.php
```diff
@@ -218,7 +218,7 @@ public function next_frame($timeout = 0)
...
      * @param string $args
```
postalservice14: While you're at it, looks like `$args` is most commonly of type `AMQPWriter`.  Maybe change this to `AMQPWriter|string`

PhpAmqpLib/Channel/AbstractChannel.php
```diff
@@ -347,6 +347,10 @@ public function wait($allowed_methods = null, $non_blocking = false, $timeout =
         }
     }
 
+    /**
+     * @param string $allowed_methods
```
postalservice14: `$allowed_methods` is actually an array

postalservice14:
@rcatlin Thanks for doing this, the docblocks desperately need to be checked and fixed in many many cases.  Looks like quite a few of the ones you had put down may not be accurate though.  Could you go through them and double check them?  What I was doing was putting a breakpoint at the first line of a method and debugging the tests to see what attribute types were being passed in.  That's a good start to figure out what the types are...

Thanks again for doing this... :-)

rcatlin:
@postalservice14 Not a problem! I'll go through them again and make the changes you've suggested.  I'll  ping you when it's ready for review again. It would be great to have all the doc blocks setup.

PhpAmqpLib/Channel/AbstractChannel.php
rcatlin: Correct. Will fix! I also agree with giving it a more descriptive name.

PhpAmqpLib/Channel/AbstractChannel.php
rcatlin: It must be a string.
`AbstractChannel#dispatch()` passes `$method_sig` into `AbstractChannel#$methodMap` methods. 

See https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Helper/Protocol/MethodMap091.php specifically the `valid_method` and `get_method`.

Also `$method_sig` is "built" using `AbstractChannel#build_method_signature()` which returns a string.

PhpAmqpLib/Channel/AbstractChannel.php
rcatlin: Yes, it is! I'll fix.

PhpAmqpLib/Channel/AbstractChannel.php
rcatlin: Got it!

PhpAmqpLib/Channel/AbstractChannel.php
rcatlin: A `$method_sig` argument is an array in these methods:

`AbstractChannel#send_method_frame()`
`AbstractConnection#prepare_channel_method_frame()`
`AbstractConnection#send_channel_method_frame()`

A `$method_sig` argument is a string in these methods:

`AbstractChannel#dispatch()`

I've changed the appropriate doc blocks to reflect these types.

rcatlin:
@postalservice14 Ready for review!

rcatlin:
Also, I would consider using the library https://github.com/beberlei/assert for asserting the correct primitive and type arrays are being used. It is also a great preparation for strict typing in PHP 7.

postalservice14:
@rcatlin Thanks for pointing out the problems with the generated classes in this PR.  I went ahead and updated those as well as the generator `spec/parser.php`, so those PHPDoc types won't get messed up again.  You'll have to rebase your changes off of master again.

One question, a few different places you have a variable named `$amqMessage` rather than `$amqpMessage`.  Was there a reason for dropping the "p"?

Other than that, looks great.

rcatlin:
@postalservice14 Thanks for the update! I'll rebase off master and change my misnamed `$amqMessage` variables to `$amqpMessage`.

rcatlin:
@postalservice14 
I've made the changes you've request and this PR is ready for code review again.

Please let me know if there's anything else I can do!

PhpAmqpLib/Helper/Protocol/MethodMap080.php
```diff
@@ -116,7 +116,7 @@ public function get_method($method_sig)
 
     /**
      * @var string $method_sig
-     * @return bool
+     * @return boolean
```
postalservice14: You don't want to change the generated classes manually.  These are generated from the `spec/parser.php`.  Also, I think we want to keep `bool`, not `boolean`.

PhpAmqpLib/Channel/AMQPChannel.php
```diff
@@ -644,7 +644,7 @@ public function queue_declare(
      * Confirms a queue definition
      *
      * @param AMQPReader $reader
-     * @return string[]
+     * @return array
```
postalservice14: Actually `string[]` was correct.

PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -894,7 +900,7 @@ public function set_connection_unblock_handler($callback)
     /**
      * Gets the connection status
      *
-     * @return bool
+     * @return boolean
```
postalservice14: Let's go with `bool` not `boolean` for consistency.

PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -932,7 +938,7 @@ protected function closeChannels()
     /**
      * Should the connection be attempted during construction?
      *
-     * @return bool
+     * @return boolean
```
postalservice14: Let's go with `bool` not `boolean` for consistency.

PhpAmqpLib/Helper/Protocol/MethodMap091.php
```diff
@@ -87,7 +87,7 @@ public function get_method($method_sig)
 
     /**
      * @var string $method_sig
-     * @return bool
+     * @return boolean
```
postalservice14: This is also a generated class.

PhpAmqpLib/Helper/Protocol/Protocol080.php
```diff
@@ -1238,7 +1238,6 @@ public static function testTableOk(AMQPReader $reader)
     }
 
     /**
-
```
postalservice14: This is also a generated class.

postalservice14:
I went ahead and made those minor changes.  Thanks so much!

PR #376, opened 2016-02-21
Verify read_write_timeout is at least 2x the heartbeat (if set)

postalservice14:


skolodyazhnyy:
@postalservice14 can you explain this "Verify read_write_timeout is at least 2x the heartbeat (if set)"?

Issue #378, opened 2016-03-01
StreamIO doesn't call restore_error_handler when an error occurs


gcalore:
As the title says, in the `StreamIO::connect` method, if something wrong happens when calling `stream_socket_client`, an exception is thrown and the `restore_error_handler` is never called.

This means that if there's a problem when connecting, and I don't want to terminate the application but continue with the normal flow of things, the custom error handler will pick up anything from that point onwards instead of the error handlers you might have defined in your application.

Luckily `AMQPSocketConnection` doesn't use custom error handlers, so I ended up using that.

I have no idea atm of what could be the solution to this problem, but I thought at least to report it.

postalservice14:
As far as I can tell, it is being called.  If the custom error handler returns null, this one gets called: https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Wire/IO/StreamIO.php#L132

If the custom error handler doesn't return null and throws an exception, this one get's called: https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Wire/IO/StreamIO.php#L318

Am I missing something?

gcalore:
Maybe I am missing something. I'm pretty sure the library handler was picking up errors from the application I'm working on, but maybe something else is going on. I'll double-check later and let you know.

postalservice14:
Are you using the latest version of php-amqplib?

gcalore:
I'm using version 2.6.1, on PHP 5.6.18.

I've been looking at this again, I didn't notice the call inside the error_handler, my bad. But the problem is still there.

I've been debugging this with invalid connection parameters (I simply turned off my rabbitmq server), and I see that the restore call inside the handler is executed as it should be, but somehow it still goes inside the error handler again during the same request with a completely unrelated error. It's like the restore function is completely ignored... my impression is that it's some odd php behaviour.

I wrote some code to test this outside my application, just in case:
https://gist.github.com/gcalore/3c5c67e140734ee89e5c

Try running `test.php` with a running RabbitMQ server that accepts the credentials set in the file, and then run it with the server turned off.

postalservice14:
Interesting, looks like the `restore_error_handler` isn't working when called from the error handler itself.  I have a fix.  Just running it through some tests now...  Thanks for the catch.

postalservice14:
PR is open, any chance you can make sure it fixes your issue? https://github.com/php-amqplib/php-amqplib/pull/379

[...]

PR #380, opened 2016-03-06
Error handler issue

gcalore:
This is a follow-up of #378 and #379. The commit messages should be enough to explain the bugfix.
The only problem is that I don't really know how to test the other three places where the error handler is used: I have no idea how to mock php core functions, and I can't think of a way to reproduce the potential errors. Any suggestion is welcome.

### DIFF
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -129,7 +129,7 @@ class StreamIO extends AbstractIO
                 STREAM_CLIENT_CONNECT,
                 $this->context
             );
-        } catch (\Exception $e) {
+        } catch (\ErrorException $e) {
             restore_error_handler();
             throw $e;
         }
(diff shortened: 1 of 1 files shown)


drealecs:
I confirm this is the way to fix the problem. :+1:

I was testing some reliability reconnect in case of connection drops by closing the TCP connection from windows and since 2.6.0 this problem appeared and it was bothering me today again.

I found the problem and did an almost identical fix for it :) : https://github.com/php-amqplib/php-amqplib/compare/master...drealecs:fix_error_handler_restore
When I tried to open a pull request I checked existing one first and I guess now it doesn't make sense to do another one.

I am able to replicate this and my conclusion is that the restore_error_handler was called two times, restoring one of the other handler; in my case restoring the one around fwrite.

Unfortunately, I don't think it's feasible to do an automatic test.
I am manually closing TCP connections using TCPview while running a complex test script that starts multiple producers and consumers and compare their outcome to ensure "at least once" delivery.
While my test is also linked with how RabbitMQ is configured and it is too much for the need here, I think it's easy to replicate by making a loop with a simple command like `queue_declare` and dropping the TCP connection.

Let me know how I could help.

postalservice14:
Thank you!

PR #381, opened 2016-03-18
Test if array_push() argument 1 is an array

sp72:
Sometimes this value is NULL and triggers an E_WARNING.

### DIFF
--- a/PhpAmqpLib/Connection/AbstractConnection.php
+++ b/PhpAmqpLib/Connection/AbstractConnection.php
@@ -575,9 +575,10 @@ class AbstractConnection extends AbstractChannel
                 //for later, when the other channel is looking for frames.
                 // Make sure the channel still exists, it could have been
                 // closed by a previous Exception.
-                if (isset($this->channels[$frame_channel]->frame_queue) && is_array($this->channels[$frame_channel]->frame_queue)) {
-                    array_push($this->channels[$frame_channel]->frame_queue, array($frame_type, $payload));
+                if (!is_array($this->channels[$frame_channel]->frame_queue)) {
+                    $this->debug->debug_msg('Frame queue is not array, type: ' . gettype($this->channels[$frame_channel]->frame_queue));
                 }
+                array_push($this->channels[$frame_channel]->frame_queue, array($frame_type, $payload));
 
                 // If we just queued up a method for channel 0 (the Connection
                 // itself) it's probably a close method in reaction to some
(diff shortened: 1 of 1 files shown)


postalservice14:
Are you able to write a test that covers this scenario?

sp72:
Sorry, I can't write a precise test. I saw this warning in my production logs and I wanted to avoid the E_WARNING. I did not have the context with the error.

postalservice14:
What is the warning you are getting?  `isset` should be checking if `$this->channels[$frame_channel]` is null.  I just want to make sure we're aren't adding needless complexity to this section of the code.

sp72:
I had the warning below which occurred because `$this->channels[$frame_channel]->frame_queue` was not an array at that time.

> Warning: array_push(): Argument #1 is not an array

I got this error a few times, something went wrong somewhere before this line and died here.
The weakness of this block is to assume that the argument #1 is always an array.
My fix checks this point. Maybe, I should throw an exception instead of doing nothing if the expected array is not one ?

videlalvaro:
I think this fix is masking a different issue, which is assuming that the channel here `$this->channels[$frame_channel]` still exists. So I would suggest fixing that other issue instead of just testing for array, since that's the effect of the other problem

postalservice14:
I agree with @videlalvaro in that I'd like to get to the bottom of the root issue first.  Are you able to put some debugging code in there and tell us the type of `$this->channels[$frame_channel]->frame_queue` when you get that warning?

Something like this?
```php
if (!is_array($this->channels[$frame_channel]->frame_queue)) {
    $this->debug->debug_msg('Frame queue is not array, type: ' . gettype($this->channels[$frame_channel]->frame_queue));
}
```

What's weird is I don't see how `frame_queue` could ever not be an array.  It's initialized as an array, and then never changed anywhere in the code to anything but an array.

What version of the library are you using?

sp72:
We use `"videlalvaro/php-amqplib": "2.5.*"`.
I understand your point about masking a different issue, adding a debug log can be a clue if the problem comes back. I updated the code with your suggestion.
With my team, we will check if an upgrade to `2.6.*` can be done in our project.

postalservice14:
@sp72 If you are using 2.5.* that is probably what the problem is...  The `isset` check you are looking at wasn't added until 2.6.*.  Let me know if simply upgrading to 2.6.* fixes this issue (it should).

postalservice14:
Closing this for now.  Please re-open if upgrading to 2.6 didn't fix the issue.

Thanks!

PR #392, opened 2016-05-02
Remove large "fixtures" from tests folder

IMSoP:
The functional tests currently contain a couple of files of (presumably random) data for exercising a RabbitMQ server, which significantly bump the size of the repository. The 4MB file is completely unused, and 1MB file can easily be regenerated on the fly.

Unfortunately, clones of the git repo will still include these files in the history (I don't know any way to prevent that), but at least this way exports or shallow clones needn't.


postalservice14:
Thanks!

PR #396, opened 2016-05-05
Adding @throws annotations

cgardner:


postalservice14:
So I'm up for a discussion on this, but for me...

I don't like listing every possible exception that could be thrown in a method.  It get's really muddy.  What this really shines light on is how we aren't catching all these exceptions and throwing more accurate specific exceptions or dealing with them in a different way.

An example of what I mean is that yes, today, `AMQPChannel::channel_alert()` method can throw six different types of exceptions, but that is shining light on upstream methods not properly catching them and throwing specific exceptions.  If `read_short()` has an exception, I would expect it to throw a `ReadException` or something like that.  Then `AMQPChannel::channel_alert()` only has to worry about one type of exception, and it can react to it.  Whether be throwing it up, or creating a more specific exception `AMQPChannelException`.

All that to say, thank you for doing this work, but rather than add all these throws, I'd rather spend some time on the exception handling of the project.

Again, open to other thoughts...

cgardner:
I agree. I wanted to just expose things as they are at the moment.  I'll
continue on this and get the exceptions cleaned up.

michaelklishin:
Looks like there was no consensus reached and some of the added annotations document very generic exceptions => closing.

PR #399, opened 2016-05-11
add message delivery tag for internal ack handler

prolic:


prolic:
ping @postalservice14 @nubeiro

prolic:
ping @michaelklishin

prolic:
hello??? any feedback on this?

hairyhum:
@prolic can you please provide more context for this PR? And also some tests that will verify that this change is necessary?

michaelklishin:
This seems to try to attach a delivery tag sent by RabbitMQ (on a channel w/ enabled publisher confirms). I'm afraid this is confusing as delivery tags on messages are typically only set for deliveries.

Please clarify what the goal here and we'll consider re-opening this but at first glance this looks like something I wouldn't accept in any client libraries I maintain.

prolic:
@michaelklishin I was working on an unit test for this, until I get notice you closed it already.

I use this for publisher confirms, correct.

The reason is, that if the delivery tag is missing (which is set by RabbitMQ), I cannot use this information in the nack handler.

It would also be consistent behaviour with php-amqp extension, see here: https://github.com/pdezwart/php-amqp/blob/master/stubs/AMQPChannel.php#L197-L216

Having the delivery tag available, allows to give specific nack handlers to tell message with delivery tag XY has been rejected, instead of only referring to the message  body (which could be the same for different messages).

prolic:
@michaelklishin any comments on my last comment?

michaelklishin:
@prolic I get notifications for the repository, no need to ping me personally every time.

prolic:
sure, will do

prolic:
added tests!

michaelklishin:
The tests don't seem to pass on PHP 5.3 through 5.5. I restarted the build to be more certain that it's not a timing issue in the test suite.

prolic:
Ok. Will check again later.

prolic:
ok should be good now, sorry for abusing travis for this, but I don't have any EOL php versions on my system any more :-)

michaelklishin:
@prolic thank you!

prolic:
When can I expect a new release?

michaelklishin:
@prolic see #469.

PR #400, opened 2016-05-11
Redirect debug output to `defined` destination

thesebas:
Allow redirecting debug output to somewhere else than STDOUT - like STDERR or any other stream/file (like Monolog stream wrapper)

### DIFF
--- a/PhpAmqpLib/Helper/DebugHelper.php
+++ b/PhpAmqpLib/Helper/DebugHelper.php
@@ -7,7 +7,12 @@ class DebugHelper
      * @var bool
      */
     protected $debug;
-
+    
+    /**
+     * @var resource
+     */
+    protected $debug_output;
+    
     /**
      * @var string
      */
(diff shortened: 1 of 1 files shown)


michaelklishin:
This makes a lot of sense. @nubeiro @hairyhum any objections to merging it?

PhpAmqpLib/Helper/DebugHelper.php
```diff
@@ -101,6 +107,6 @@ public function debug_connection_start($version_major, $version_minor, $server_p
      * @param string $s
      */
     protected function print_msg($s) {
-        echo $s . PHP_EOL;
+        fprintf($this->debug_output, $s . PHP_EOL);
```
hairyhum: `fprintf` expects a format string and arguments to be provided, so the semantics is different from `echo` (in presence of `%` symbol). Should it be a `fwrite` instead?

thesebas: I agree, `fwrite` is better here.

thiagotalma:
Hi guys!

This change triggered this exception:
`Use of undefined constant STDOUT - assumed 'STDOUT'`

@thesebas @michaelklishin @nubeiro

michaelklishin:
@thiagotalma that's a little light on details. In what environment, with what kind of configuration? Feel free to submit a PR that sets the default to be standard output that'd avoid warnings and such.

thiagotalma:
https://github.com/php-amqplib/php-amqplib/pull/501  ;)

PR #401, opened 2016-05-11
Re-enable heartbeats after reconnection

bmcgavin:
I found that if a consumer with client heartbeats enabled loses and re-establishes the connection it will not send client heartbeats, causing the server to eventually force a disconnect (I am using AMQPStreamConnection / StreamIO).

This is because StreamIO::disableHeartbeat is called on close(), setting StreamIO::heartbeat to 0, disabling the sending of client heartbeats.

However the matching value AMQPStreamConnection::heartbeat is not altered, so the connection.tune / connection.tune_ok messages still define a client heartbeat, so the server is expecting one.

This PR adds a StreamIO::reenableHeartbeat method which is called after successful reconnection and sets the StreamIO::heartbeat value back to what it was set to in the constructor.

### DIFF
--- a/PhpAmqpLib/Connection/AbstractConnection.php
+++ b/PhpAmqpLib/Connection/AbstractConnection.php
@@ -215,6 +215,11 @@ class AbstractConnection extends AbstractChannel
 
                 $host = $this->x_open($this->vhost, '', $this->insist);
                 if (!$host) {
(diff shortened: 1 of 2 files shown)


Fogush:
Encountered the same issue. It would be nice to have this fix :+1:

PR #402, opened 2016-05-13
catch Throwable

fprochazka:
Half of those cases could be solved with `finally`, but since this lib supports 5.3 it has to be done like this.

Also, if at some places it doesn't make sense to catch `Throwable`, let me know, I'll change it. Makes sense to me to add it to all the places in this PR.

### DIFF
--- a/PhpAmqpLib/Connection/AbstractConnection.php
+++ b/PhpAmqpLib/Connection/AbstractConnection.php
@@ -230,6 +230,11 @@ class AbstractConnection extends AbstractChannel
             $this->setIsConnected(false);
             $this->closeChannels();
             throw $e; // Rethrow exception
+        } catch (\Throwable $e) {
+            // Something went wrong, set the connection status
+            $this->setIsConnected(false);
+            $this->closeChannels();
+            throw $e; // Rethrow exception
         }
     }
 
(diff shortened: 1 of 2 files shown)


PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -273,6 +278,8 @@ protected function safeClose()
             }
         } catch (\Exception $e) {
             // Nothing here
+        } catch (\Throwable $e) {
+            // Nothing here
```
fprochazka: not sure about this one

PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -931,6 +938,8 @@ protected function closeChannels()
                 $channel->close();
             } catch (\Exception $e) {
                 /* Ignore closing errors */
+            } catch (\Throwable $e) {
+                /* Ignore closing errors */
```
fprochazka: not sure about this one

michaelklishin:
@nubeiro expressed interest in dropping support for 5.3 relatively soon and there is no consensus around this change => closing.

fprochazka:
@michaelklishin catching `Throwable` does not mean droping support for `5.3`

michaelklishin:
@fprochazka I was referring to

> since this lib supports 5.3 it has to be done like this

If we are going to drop 5.3 support then it can be solved in a different (and presumably more sensible) way.

fprochazka:
@michaelklishin Oh I see now, I've skimmed over your previous message, sorry. Alrighty then :)

michaelklishin:
@fprochazka by all means, feel free to submit a PR that addresses this differently ;)

PR #410, opened 2016-06-08
Add extensions requirements to Readme.md

jeff-innovent:
Document that the lib requires the bcmath and mbstring extensions.


michaelklishin:
Thank you!

PR #417, opened 2016-07-01
Fix wrong date when maintainers changed.

### LINKED ISSUE #411: Readme
Message from future?

https://github.com/php-amqplib/php-amqplib#new-maintainers-2016-10-02

dopesong:
Closes #411


michaelklishin:
Consider using "Feb 10th" instead of the format you have. Your PR simply flips the problem on its back.

PR #435, opened 2016-09-06
Reduce memory usage in StreamIO::write()

vojtech-dobes:
It seems that PHP 7 behaves differently than PHP 5.\* - original mb_substr's output exists side-by-side with the original value of $data variable, which leads to doubling used memory. The bigger the $data is, the worse it is.


michaelklishin:
Merging per user report in #456.

vojtech-dobes:
Thanks for merging @michaelklishin! I've noticed this library is pretty quiet regarding development for some time - is this sign of changes to come :) ?

michaelklishin:
I don't know, more likely it's a sign of public holidays and the primary maintainer having other obligations.

PR #445, opened 2016-11-02
Increased overall test coverage.

nurolopher:
Increased overall test coverage by writing some unit tests, even if they were too simple. Simplified, removed unnecessary comments, codes.

### DIFF
--- a/tests/Unit/Wire/IO/SocketIOTest.php
+++ b/tests/Unit/Wire/IO/SocketIOTest.php
@@ -1,0 +1,51 @@
+<?php
+namespace PhpAmqpLib\tests\Unit\Wire\IO;
+
+use PhpAmqpLib\Wire\IO\SocketIO;
+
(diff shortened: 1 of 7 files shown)


tests/Unit/Wire/IO/SocketIOTest.php
```diff
@@ -0,0 +1,51 @@
+<?php
+namespace PhpAmqpLib\tests\Unit\Wire\IO;
+
+use PhpAmqpLib\Wire\IO\SocketIO;
+
+class SocketIOTest extends \PHPUnit_Framework_TestCase
```
mente: that looks more like functional because requires connection

michaelklishin:
@nurolopher thank you so much! Can you please rebase this against master so we can take a closer look?

nubeiro:
Thank you @nurolopher, adding tests to this code base is not straightforward, I appreciate the effort.

Issue #272, opened 2015-05-18
Error reading data. Received 0 instead of expected 7 bytes


arogachev:
I'm using Vagrant + puphpet, together with this extension. Recently I updated to the latest versions and now struggling with this issue.

There are some problems with applying configuration (at least for me).
User `guest` was existing by default, but when I tried:

```
sudo rabbitmqctl list_user_permissions guest
```

it says:

```
Listing permissions for user "guest" ...
Error: {no_such_user,<<"guest">>}
```

So I executed these commands manually for testing purposes.

```
sudo rabbitmqctl add_vhost /
sudo rabbitmqctl add_user guest guest // was already existing as default
sudo rabbitmqctl set_permissions -p / guest ".*" ".*" ".*"
```

Now if I execute:

```
sudo rabbitmqctl list_user_permissions guest
```

It gives the following results:

```
Listing permissions for user "guest" ...
/      .*      .*      .*
...done.
```

Then I restarted rabbit server like that:

```
sudo service rabbitmq-server restart
```

Connection is created like that:

```php
$connection = new AMQPConnection('127.0.0.1', 5672, 'guest', 'guest', '/');
```

According to trace, error appears right on this line. Here is full error text:

```
PhpAmqpLib\Exception\AMQPRuntimeException

Error reading data. Received 0 instead of expected 7 bytes
```
    
`puphpet` config rabbitmq section:

```yml
rabbitmq:
    install: '1'
    settings:
        port: '5672'
    users:
        user-1:
            admin: true
            name: admin
            password: admin
            permissions:
                permissions-1:
                    host: /
                    configure_permission: '.*'
                    read_permission: '.*'
                    write_permission: '.*'
    vhosts:
        - /
    plugins: {  }
```

I also tried with `admin` user, the result is the same error.

thiagotalma:
You read [this](https://github.com/videlalvaro/php-amqplib/issues/72)?

arogachev:
Yes, but seems like credentials are right.

arogachev:
Solved. This was because of old version of RabbitMQ. Updated to the last stable version - and this error disappeared.

lerox:
We had the same exception when consuming... (Debian)

Fixed with:

```diff
-use PhpAmqpLib\Connection\AMQPSocketConnection;
+use PhpAmqpLib\Connection\AMQPStreamConnection;
```

:stuck_out_tongue:

genepeng:
[...]

PR #448, opened 2016-11-07
bring heartbeat support to socket connection

mente:
As an answer for #272 where `AMQPSocketConnection` was an issue.

Actual heartbeat code is taken from `StreamIO` with adaptation to sockets.
Also includes splitting of receive timeout and send timeout as it's done in `StreamIO`, though there it's `$connect_timeout` and `$read_write_timeout` which is hard to be accomplished for sockets

As soon as this PR is merged I need to update bundle as it passes options incorrectly (or better mimic same constructor everywhere? or better have a builder to fade out all discrepancies between different connection types).

### DIFF
--- a/PhpAmqpLib/Connection/AMQPSocketConnection.php
+++ b/PhpAmqpLib/Connection/AMQPSocketConnection.php
@@ -6,19 +6,19 @@ use PhpAmqpLib\Wire\IO\SocketIO;
 class AMQPSocketConnection extends AbstractConnection
 {
     /**
-     * @param string    $host
-     * @param int       $port
-     * @param string    $user
-     * @param string    $password
-     * @param string    $vhost
-     * @param bool      $insist
-     * @param string    $login_method
-     * @param null      $login_response
-     * @param string    $locale
-     * @param float|int $timeout
-     * @param bool      $keepalive
-     * @param int       $read_write_timeout
-     * @param int       $heartbeat
+     * @param string $host
+     * @param int $port
+     * @param string $user
+     * @param string $password
+     * @param string $vhost
+     * @param bool $insist
+     * @param string $login_method
+     * @param null $login_response
+     * @param string $locale
+     * @param float $read_timeout
+     * @param bool $keepalive
+     * @param int $write_timeout
+     * @param int $heartbeat
      */
     public function __construct(
         $host,
@@ -30,12 +30,12 @@ class AMQPSocketConnection extends AbstractConnection
         $login_method = 'AMQPLAIN',
         $login_response = null,
         $locale = 'en_US',
-        $timeout = 3,
+        $read_timeout = 3,
         $keepalive = false,
-        $read_write_timeout = 3,
+        $write_timeout = 3,
         $heartbeat = 0
     ) {
-        $io = new SocketIO($host, $port, $read_write_timeout, $keepalive, $timeout, $heartbeat);
+        $io = new SocketIO($host, $port, $read_timeout, $keepalive, $write_timeout, $heartbeat);
 
         parent::__construct(
             $user,
--- a/PhpAmqpLib/Connection/AMQPSocketConnection.php
+++ b/PhpAmqpLib/Connection/AMQPSocketConnection.php
@@ -16,8 +16,8 @@ class AMQPSocketConnection extends AbstractConnection
      * @param null      $login_response
      * @param string    $locale
      * @param float|int $timeout
-     * @param int       $read_write_timeout
      * @param bool      $keepalive
+     * @param int       $read_write_timeout
      * @param int       $heartbeat
      */
     public function __construct(
@@ -31,11 +31,11 @@ class AMQPSocketConnection extends AbstractConnection
         $login_response = null,
         $locale = 'en_US',
         $timeout = 3,
-        $read_write_timeout = 3,
         $keepalive = false,
+        $read_write_timeout = 3,
         $heartbeat = 0
     ) {
-        $io = new SocketIO($host, $port, $timeout, $read_write_timeout, $heartbeat, $keepalive);
+        $io = new SocketIO($host, $port, $timeout, $keepalive, $read_write_timeout, $heartbeat);
 
         parent::__construct(
             $user,
--- a/PhpAmqpLib/Connection/AMQPSocketConnection.php
+++ b/PhpAmqpLib/Connection/AMQPSocketConnection.php
@@ -35,7 +35,7 @@ class AMQPSocketConnection extends AbstractConnection
         $read_write_timeout = 3,
         $heartbeat = 0
     ) {
-        $io = new SocketIO($host, $port, $timeout, $keepalive, $read_write_timeout, $heartbeat);
+        $io = new SocketIO($host, $port, $read_write_timeout, $keepalive, $timeout, $heartbeat);
 
         parent::__construct(
             $user,
--- a/PhpAmqpLib/Connection/AMQPSocketConnection.php
+++ b/PhpAmqpLib/Connection/AMQPSocketConnection.php
@@ -6,17 +6,19 @@ use PhpAmqpLib\Wire\IO\SocketIO;
 class AMQPSocketConnection extends AbstractConnection
 {
     /**
-     * @param string $host
-     * @param int $port
-     * @param string $user
-     * @param string $password
-     * @param string $vhost
-     * @param bool $insist
-     * @param string $login_method
-     * @param null $login_response
-     * @param string $locale
-     * @param float $timeout
-     * @param bool $keepalive
+     * @param string    $host
+     * @param int       $port
+     * @param string    $user
+     * @param string    $password
+     * @param string    $vhost
+     * @param bool      $insist
+     * @param string    $login_method
+     * @param null      $login_response
+     * @param string    $locale
+     * @param float|int $timeout
+     * @param int       $read_write_timeout
+     * @param bool      $keepalive
+     * @param int       $heartbeat
      */
     public function __construct(
         $host,
@@ -29,9 +31,11 @@ class AMQPSocketConnection extends AbstractConnection
         $login_response = null,
         $locale = 'en_US',
         $timeout = 3,
-        $keepalive = false
+        $read_write_timeout = 3,
+        $keepalive = false,
+        $heartbeat = 0
     ) {
-        $io = new SocketIO($host, $port, $timeout, $keepalive);
+        $io = new SocketIO($host, $port, $timeout, $read_write_timeout, $heartbeat, $keepalive);
 
         parent::__construct(
             $user,
@@ -41,7 +45,8 @@ class AMQPSocketConnection extends AbstractConnection
             $login_method,
             $login_response,
             $locale,
-            $io
+            $io,
+            $heartbeat
         );
     }
 }
--- a/PhpAmqpLib/Wire/IO/SocketIO.php
+++ b/PhpAmqpLib/Wire/IO/SocketIO.php
@@ -38,12 +38,12 @@ class SocketIO extends AbstractIO
     /**
      * @param string     $host
      * @param int        $port
-     * @param float      $send_timeout
-     * @param bool       $keepalive
      * @param float|null $read_timeout if null - defaults to $send_timeout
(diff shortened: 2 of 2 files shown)


mente:
Anywhere I can help to move this PR forward?

mente:
I won't give up...

mente:
What prevents this PR from being merged? 

Code - ✓
Description - ✓
Tests - ✓

What else is needed?

michaelklishin:
@mente I'm not really a PHP user so I only dare to merge small changes. Let's see what our maintainers think: @postalservice14, @nubeiro.

In case there's still no response in a few days I'll try to review it to the best of my abilities.

PhpAmqpLib/Connection/AMQPSocketConnection.php
```diff
@@ -29,9 +31,11 @@ public function __construct(
         $login_response = null,
         $locale = 'en_US',
         $timeout = 3,
-        $keepalive = false
+        $read_write_timeout = 3,
+        $keepalive = false,
+        $heartbeat = 0
     ) {
-        $io = new SocketIO($host, $port, $timeout, $keepalive);
+        $io = new SocketIO($host, $port, $timeout, $read_write_timeout, $heartbeat, $keepalive);
```
nubeiro: Any reason to introduce new parameters before old parameter?

PhpAmqpLib/Wire/IO/SocketIO.php
```diff
@@ -97,6 +115,8 @@ public function read($n)
 
         $buf = socket_read($this->sock, $n);
         while ($read < $n && $buf !== '' && $buf !== false) {
+            $this->check_heartbeat();
+
```
nubeiro: Why the extra line?

PhpAmqpLib/Wire/IO/SocketIO.php
```diff
@@ -194,4 +219,40 @@ protected function enable_keepalive()
...
+            }
+        }
+    }
+
+    /**
+     * Sends a heartbeat message
+     */
+    protected function write_heartbeat()
+    {
+        $pkt = new AMQPWriter();
```
nubeiro: Is this the only way we can get an AMQPWriter here?

PhpAmqpLib/Connection/AMQPSocketConnection.php
mente: To have similar signature to `SocketIO`

PhpAmqpLib/Wire/IO/SocketIO.php
mente: to split logical parts

PhpAmqpLib/Wire/IO/SocketIO.php
mente: Sorry, what do you mean?

skafandri: He means to inject it. Plus, I would not create a new object for each heart beat, just to give more work to the GC.

mente: `AMQPWriter` is a stateful object and there's no way to reset it. 

This part of the code was copied from `StreamIO`. This is not the best apologize but I don't see how else to write a packet. Unless doing it manually which is IMHO worse than "give more work to the GC".

What do you think?

PhpAmqpLib/Connection/AMQPSocketConnection.php
drealecs: Unfortunately this is a BC break. It should only be added in a new major version.

mente:
I've reverted BC changes

PhpAmqpLib/Wire/IO/SocketIO.php
```diff
@@ -23,16 +36,20 @@ class SocketIO extends AbstractIO
...
-     * @param bool $keepalive
+     * @param string     $host
+     * @param int        $port
+     * @param float      $send_timeout
+     * @param bool       $keepalive
+     * @param float|null $read_timeout if null - defaults to $send_timeout
+     * @param int        $heartbeat
      */
-    public function __construct($host, $port, $timeout, $keepalive = false)
+    public function __construct($host, $port, $send_timeout, $keepalive = false, $read_timeout = null, $heartbeat)
```
drealecs: The third parameter `$timeout` was used as a read timeout and now it's changed to write timeout.
I believe you need to switch the parameter in the method's signature.

PhpAmqpLib/Connection/AMQPSocketConnection.php
```diff
@@ -6,17 +6,19 @@
...
-     * @param string $password
-     * @param string $vhost
-     * @param bool $insist
-     * @param string $login_method
-     * @param null $login_response
-     * @param string $locale
-     * @param float $timeout
-     * @param bool $keepalive
+     * @param string    $host
+     * @param int       $port
```
drealecs: Don't know all the rules related to format for this repo but personally I prefer not to change the format of lines I don't touch. This way, the diff would have been cleaner and also the blame wouldn't by formatting stuff.
I suggest reverting the format to the original one, without aligning variable names.

PhpAmqpLib/Connection/AMQPSocketConnection.php
```diff
@@ -29,9 +31,11 @@ public function __construct(
         $login_response = null,
         $locale = 'en_US',
         $timeout = 3,
```
drealecs: Naming could be better I believe.
`$timeout` -> `$read_timeout`
`$read_write_timeout` -> `$write_timeout`

mente:
@drealecs thanks for your comments! Hope I've fixed'em all :)

PhpAmqpLib/Wire/IO/SocketIO.php
```diff
@@ -25,14 +38,18 @@ class SocketIO extends AbstractIO
...
      * @param string $host
      * @param int $port
-     * @param float $timeout
+     * @param float $read_timeout
      * @param bool $keepalive
+     * @param float|null $write_timeout if null defaults to read timeout
+     * @param int $heartbeat
      */
-    public function __construct($host, $port, $timeout, $keepalive = false)
+    public function __construct($host, $port, $read_timeout, $keepalive = false, $write_timeout = null, $heartbeat)
```
drealecs: if you could add a default for `$heartbeat` parameter also, it would be great.

mente: Not sure, heartbeat heavily depends on timeout and should be 2x. Otherwise bugs are expected. Don't want to make it buggy by default

drealecs: I meant a default value for backward compatibility. A value of 0/false/null that would mean no heartbeat.

drealecs: pinging @mente for this last part of the review

drealecs:
Hi @mente,

I believe that some changes I helped with, related to heartbeat in stream connection, should be replicated here. Either after merging this PR or in it.
Please see the changes here: https://github.com/php-amqplib/php-amqplib/pull/473

Thanks,
Alex

mente:
Hey guys, sorry about delay. I'm overwhelmed by amount of work on my daily job. I will try to check/fix it till the end of next week

PhpAmqpLib/Wire/IO/SocketIO.php
mente: made default to 0 which means heartbeat is off

mente:
> I believe that some changes I helped with, related to heartbeat in stream connection, should be replicated here. Either after merging this PR or in it.

Can you describe how to reproduce it? I bearly understood the problem tbh :-/

PhpAmqpLib/Wire/IO/SocketIO.php
```diff
@@ -41,9 +41,9 @@ class SocketIO extends AbstractIO
      * @param float $read_timeout
      * @param bool $keepalive
      * @param float|null $write_timeout if null defaults to read timeout
-     * @param int $heartbeat
+     * @param int $heartbeat how often to send heartbeat. 0 means off
      */
-    public function __construct($host, $port, $read_timeout, $keepalive = false, $write_timeout = null, $heartbeat)
+    public function __construct($host, $port, $read_timeout, $keepalive = false, $write_timeout = null, $heartbeat = 0)
```
michaelklishin: This is a terrible idea. Please use the same default value as other clients use, which these days is 60 seconds.

michaelklishin:
Heartbeats should not be disabled by default.

PhpAmqpLib/Wire/IO/SocketIO.php
mente: Shouldn't heartbeat by default be 2x of read timeout?

michaelklishin: @mente it depends on the implementation. Maybe in this one it should be, and it's a good approximation. My point is that it should be enabled by default, not that a particular value is the right one.

drealecs: @michaelklishin 
I believe this is a BC break as the functionality will change when upgrading.
Unfortunately, even if the recommended default should be to have heartbeat enabled and set to 60, for this PHP extension the default was to have it disabled.
I believe the original authors did this because it is not very easy to handle heartbeat logic in PHP, or at least it's not easy when not using coroutines and async behaviours.

I believe it is not safe to turn on this in a minor version as it can break library usage.

drealecs:
I think the best is to do it in a separate pull request after merging this
one to have small changes.

drealecs:
> Heartbeats should not be disabled by default.

Unfortunately, even if the recommended default should be to have heartbeat enabled and set to 60, for this PHP extension the default was to have it disabled.
I believe the original authors did this because it is not very easy to handle heartbeat logic in PHP, or at least it's not easy when not using coroutines and async behaviours.

I believe it is not safe to turn on this in a minor version as it can break library usage.

PhpAmqpLib/Wire/IO/SocketIO.php
michaelklishin: @drealecs so what? We can bump the version for the next release to `2.7.0` if needed.

michaelklishin:
@drealecs I wouldn't object to having heartbeats enabled by default in a separate PR.

drealecs:
Sorry, my reply was unclear to whom it replies.
IMO, I think enabling heartbeats should be done on a new major version to avoid BC changes in behavior.

I think version 3.0.0 is a good place to contain this.

We could open a new pull requests for it but before that maybe you can define what branch will be for v3. Should it be master and you will create a new master-v2? or a new branch for v3?

mente:
So did we agree on something? Should I enable heartbeats by default or leave as it is to preserve backwards compatibility?

michaelklishin:
@mente leave it as is and submit another PR that enables them. We will get to it at a later point.

mente:
Does it mean this PR won't be merged until I create new one?

drealecs:
@michaelklishin your current review status is "Changes requested". Maybe you can change that as you agreed to make the change separately.
From my point of view, I'm fine with the current changes.

hairyhum:
@mente can you please add a cleanup code to the `close` function. The same as in `StreamIO`? The `reconnect` function can have some issues with non-clean timestamps.

hairyhum:
@michaelklishin can you please verify if your request for changes still stands?

PhpAmqpLib/Wire/IO/SocketIO.php
michaelklishin: Heartbeats are usually sent 5-20 times a minute (the interval is configurable). If that's too much GC overhead for you, you shouldn't be using a language such as PHP to begin with.

michaelklishin:
@hairyhum done. I won't merge before your request for a code cleanup is addressed then.

michaelklishin:
@mente please rebase this against master and then I think we should merge this (refactoring ideas can be submitted later).

mente:
Done

michaelklishin:
@mente thank you! Looks like we are ready to cut an RC of the next release now. @hairyhum @nubeiro

PR #459, opened 2017-01-14
Bugfix for HHVM compatability

Octember:
mb_strlen will error if passed null on HHVM, this is a workaround for that.


michaelklishin:
Merging as the change is small and 7 different configurations pass on Travis.

michaelklishin:
@Octember thank you!

Octember:
@michaelklishin thank you! This is huge for us.

Octember:
@michaelklishin Thank you! This helps a lot.

PR #460, opened 2017-01-18
Ensure integer values are actually integers for HHVM compatibility

Octember:
HHVM will error if `sec` or `usec` are null, and they are specifically passed as null here: https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Wire/IO/StreamIO.php#L227

### DIFF
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -405,6 +405,8 @@ class StreamIO extends AbstractIO
         $write = null;
         $except = null;
         $result = false;
+        $sec = is_int($sec) ? $sec : 0;
+        $usec = is_int($usec) ? $usec : 0;
 
         set_error_handler(array($this, 'error_handler'));
         try {


fsalehpour:
According to http://php.net/manual/en/function.stream-select.php:

>  If tv_sec and tv_usec are both set to 0, stream_select() will not wait for data - instead it will return immediately, indicating the current status of the streams.

This causes the `stream_select` to return `0` which causes the `AMQPTimeoutException` to be thrown in almost every case that a timeout value is used in `wait()` and other functions. 

Try running https://github.com/php-amqplib/php-amqplib/blob/master/demo/basic_qos.php

This is also the main cause for https://github.com/php-amqplib/php-amqplib/issues/462 which I filed before.

thiagotalma:
Hi guys!
This change wrecked my server.... 😭 
It must have happened to more people. I had to comment on the 2 lines in order to reactivate the consumer.

michaelklishin:
I reverted it.

thiagotalma:
Thanks!

PR #463, opened 2017-01-24
Add work-around to fix 462

fsalehpour:
Adds the necessary code to check if the script is being run under HHVM and set the timeout parameters accordingly before calling stream_select

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ php:
   - 5.5
   - 5.6
   - 7.0
+  - 7.1
   - hhvm
 
 # This triggers builds to run on the new TravisCI infrastructure.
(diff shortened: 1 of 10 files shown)


hairyhum:
For the `stream_select` function `0` and `null` are different things. `0` means "no wait", while `null` means "wait indefinitely".
According to [HHVM code](https://github.com/facebook/hhvm/blob/222ae58713f5ff44aba46206caf7a021c51b81be/hphp/runtime/ext/sockets/ext_sockets.cpp#L975) `$sec` parameter can be `null`, while `$usec` should be an integer. @fsalehpour can you please update the branch to only convert `$usec` to integer?

fsalehpour:
@hairyhum Thanks and done.

I don't run HHVM myself. So please let me know if it needs any more updates.

michaelklishin:
@fsalehpour thank you. According to GitHub this branch now has conflicts with master. Can you please rebase?

fsalehpour:
@michaelklishin done. :smile:

PR #467, opened 2017-02-19
Fix signal handling exit example

dzuelke:
Processes that terminate in response to a signal must not exit manually, but instead terminate themselves using the received signal number, so that calling programs can reliably determine the cause of the termination. This is important for example in scripted bash loops to correctly terminate on Ctrl+C.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,8 +183,8 @@ $pcntlHandler = function ($signal) {
         case \SIGUSR1:
         case \SIGINT:
             // some stuff before stop consumer e.g. delete lock etc
-            exit(0);
-            break;
+            pcntl_signal($signal, SIG_DFL); // restore handler
+            posix_kill(posix_getpid(), $signal); // kill self with signal, see https://www.cons.org/cracauer/sigint.html
         case \SIGHUP:
             // some stuff to restart consumer
             break;


michaelklishin:
This make the exit code of the value of the signal, which

 * Most developers don't know
 * It probably does not match any [decently structured exit code code map](https://linux.die.net/include/sysexits.h)

so the benefit here is in

 * Existing with a non-0 value
 * Restoring the handler

Is my reading of this correct? If so, I'd recommend exiting with either `EX_SOFTWARE` (from the link above) or `1` instead. I don't see how it would be convenient for the callers to use signal values as exit codes. FWIW this is the first time I personally see this particular convention.

michaelklishin:
I re-read the post and my understanding is that parent processes are supposed to use `WTERMSIG`  and `WIFSIGNALED` to find out if a child exited with due to a `SIGINT`, not its exit code. Therefore my suggestion of using a different exit code shouldn't interfere.

dzuelke:
Your reading is not correct :)

Any program that exits in response to a signal must not exit with a code (and definitely not with code 0), but instead kill itself using that same signal.

Only then can the OS relay the `WIFSIGNALED` information etc.

The shell displays this as 128+`$SIGNAL`. If you run `sleep 10` on your terminal, hit `Ctrl+C`, and then run `echo $?`, the result will be 130 - 128+SIGINT. If you quickly `killall sleep` from another terminal instead, the exit code will be 143 (128+SIGTERM).

This behavior cannot be emulated by an `exit(2);` or similar.

Programs not behaving this way will cause trouble e.g. for supervisor daemons, or in shell loops, as the article states.

Any Unix program behaves this way. You can try it with `sleep`, or `cat`, or whatever:

```
$ cat
^C
$ echo $?
130
```

```
$ cat # and then "killall cat" from another terminal
Terminated: 15
$ echo $?
143
```

PR #473, opened 2017-03-18
half-open connection detection with heartbeat while reapeating 

drealecs:
When consuming and waiting for a message, if server becomes suddenly unreachable (network split, machine fail) only the StreamIO select is called (possibly with a timeout) in `AMQPReader::wait()` and exists with an exception: AMQPTimeoutException that can be handled successfully and StreamIO read is never called to be able to trigger heartbeat logic.
We use this pattern with calling `AMQPChannel::wait()` with 1 second timeout while consuming in order to do some other checks and want to rely on heartbeat in this case also.
We use it in a worker and need to detect as soon as possible when the connection drops but messages might get into the queue with one hour interval.

Was thinking about adding it to the StreamIO write method also.

From what I can see, the heartbeat logic is fast and I'm not concerned about performance.

### DIFF
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -197,12 +197,12 @@ class StreamIO extends AbstractIO
      */
     public function read($len)
     {
+        $this->check_heartbeat();
+
         $read = 0;
         $data = '';
 
         while ($read < $len) {
-            $this->check_heartbeat();
-
             if (!is_resource($this->sock) || feof($this->sock)) {
                 throw new AMQPRuntimeException('Broken pipe or closed connection');
             }
@@ -401,6 +401,8 @@ class StreamIO extends AbstractIO
      */
     public function select($sec, $usec)
     {
+        $this->check_heartbeat();
+
         $read = array($this->sock);
         $write = null;
         $except = null;


drealecs:
Is there a way to speed things up a bit?
There are a lot of pending work. I'll work on reviewing the changes in the other pull requests. Let me know what else we could do.
Thanks!

michaelklishin:
@nubeiro does this look good to you?

drealecs:
Tried to think about an automated test but it's very hard to do one.

I'm testing it manually by creating two docker containers in a docker network: rabbitmq server and php client.
I start the worker on php client container and add some logging in different places in the library to see heartbeats sending and receiving.
To simulate half open connection I remove the php container from the docker network and see what happens.

I'm not sure automating this is doable or would help. Let me know what you think. Thanks.

drealecs:
Pinging again the maintainers and reviewers :)

michaelklishin:
@nubeiro @postalservice14 would you have a chance to take a look? I'm going to trust @drealecs's judgement in merge it in 2 days if we get no further comments. Thank you.

drealecs:
I have already test the changes in staging environment and I can confirm it fixed our issue.
Our new environment is running on docker swarm mode and networks connection seems to fail more often then on VMs.

Have you guys had time to take a look?

As a separate topic, can I help with any other work related to next release?
If there are some important issues, maybe you can mark them with a milestone to make things clear.
Thanks!

michaelklishin:
@drealecs thank you for the offer. Going through open pull requests, assessing and QA'ing them is by far the best way to help. I'm hardly a PHP expert but can help with any RabbitMQ or client library design question.

michaelklishin:
Merged per https://github.com/php-amqplib/php-amqplib/pull/473#issuecomment-291131162 and https://github.com/php-amqplib/php-amqplib/pull/473#issuecomment-293367856.

Issue #413, opened 2016-06-15
Dropped connections and heartbeats causing bad_header error and infinite loop


DemonTPx:
We have a connection going on between a datacenter and amazon AWS through a elastic load balancer. When trying to secure this connection using SSL, I've bumped into a problem.

When using SSL for the stream, keepalive can no longer be enabled (see issue #371). So I decided to try enabling heartbeats.

After a while I found out that a lot of my php processes just hang and after a bit more investigation, they seems to be stuck at the reconnect phase, triggered by the heartbeat functionality.

It seems to me that the following happens:
- My php script connects to the rabbitmq server
- My php script then does some stuff which does not involve rabbitmq for a while
- The server reports that the client missed a heartbeat
- My php script then wants to insert something into rabbitmq, which triggers the heartbeat, which tries to reconnect, since it thinks the server has gone away
- The rabbitmq server receives an unexpected package and closes the connection immediately
- The heartbeat is triggered again which, tries to reconnect again

The last two steps are executed indefinitely.

I tried disabling SSL, and it gives the same result.

I dug in some more and ran wireshark to figure out what was happening under the hood. When connecting for the first time, it seems that the communication goes like this:
- Server sends protocol frame
- Server sends Connection.Start frame
- Client sends Connection.Start-Ok frame
- Server sends Connection.Tune frame
- Client sends Connection.Tune-Ok frame

But when the heartbeat is triggered this happens:
- Client sends heartbeat frame
- Server sends protocol frame
- Server disconnects

The rabbitmq log looks like this:

```
=INFO REPORT==== 15-Jun-2016::09:41:56 ===
accepting AMQP connection <0.32068.24> (10.0.0.250:41049 -> 10.0.6.10:5672)

=ERROR REPORT==== 15-Jun-2016::09:41:59 ===
closing AMQP connection <0.32068.24> (10.0.0.250:41049 -> 10.0.6.10:5672):
missed heartbeats from client, timeout: 1s

=INFO REPORT==== 15-Jun-2016::09:42:12 ===
accepting AMQP connection <0.32228.24> (10.0.0.250:41063 -> 10.0.6.10:5672)

=ERROR REPORT==== 15-Jun-2016::09:42:12 ===
closing AMQP connection <0.32228.24> (10.0.0.250:41063 -> 10.0.6.10:5672):
{bad_header,<<8,0,0,0,0,0,0,0>>}
```

[...]

Issue #469, opened 2017-02-28
Is it time for a new release?


therosco:
It's been almost a year since the previous release.

Do you mind to collect all these patches and fixes happened during this time under a new shining release?

crashev:
Also wonder when will be new version. Old stopped working properly with latest 3.6.7 version of Rabbitmq

michaelklishin:
@crashev you are making claims with no evidence provided (in #472). Please stop it, collect logs and really investigate what's going on.

michaelklishin:
FTR, the issue was with Erlang/OTP R16B03 and 17.x compatibility that was broken in 3.6.7. [3.6.8 is already out](https://github.com/rabbitmq/rabbitmq-server/releases/tag/rabbitmq_v3_6_8).

crashev:
@michaelklishin Thanks for information. Everything works just fine with rabbitmq 3.6.8, sorry to bother You here without deeper checking :)

nubeiro:
@michaelklishin so we can close this issue?

therosco:
Has new version been released yet?

nubeiro:
If you refer to php-amqplib, not yet.

therosco:
Ok then, let the issue be open until php-amqplib would be released.

michaelklishin:
@nubeiro we will close it after the GA release.

FTR, @nubeiro @hairyhum and I have an RC of 2.6.4 planned very soon (a couple more PRs need a bit of massaging).

michaelklishin:
@nubeiro @hairyhum both PRs I had in mind for the next release are now merged.

Should we tag an RC? Should it be a `2.6.4` or `2.7.0`? Given the number of changes I'd suggest the latter, even if we don't technically have any breaking changes.

hairyhum:
There is a new behaviour in SocketIO, so I think it should be `2.7.0`. And there should be an RC because of this new behaviour.

michaelklishin:
Yes, there will be an RC. The question was really "should we tag it now".

michaelklishin:
@nubeiro I think we can tag a `2.7.0-rc1`. The volume of changes alone is a good enough reason to jump to `2.7.0`. Would you agree? Can you please tag it this time (you've mentioned over email that there is a certain tag name pattern that Packageist expects)?

nubeiro:
Agreed, I will tag with 2.7.0-rc1 then.

nubeiro:
[...]

PR #479, opened 2017-04-21
Fix infinite loop on reconnect - check_heartbeat

ruicampos:
The problem this PR fixes is:
* StreamConnection with heartbeat is created (e.g. heartbeat=10 seconds)
* Start consuming messages from the queue
* If one of the messages take more than 2*heartbeat interval to process (e.g. 30 seconds), the next time it tries to read something from Rabbit it will check_heartbeat()
* As it finds that it passed more than 2*heartbeat, it will reconnect()
* But as it does not reset the values of ``last_read`` and ``last_write``, after reconnect it will do the check_heartbeat() again and as it is based on ``last_read``, it will try to reconnect again
* It keeps trying to reconnect in infinite loop

This PR fixes issues: https://github.com/php-amqplib/php-amqplib/issues/413 and https://github.com/php-amqplib/php-amqplib/issues/309

### DIFF
--- a/PhpAmqpLib/Wire/IO/StreamIO.php
+++ b/PhpAmqpLib/Wire/IO/StreamIO.php
@@ -247,6 +247,8 @@ class StreamIO extends AbstractIO
             fclose($this->sock);
         }
         $this->sock = null;
+        $this->last_read = null;
+        $this->last_write = null;
     }
 
     /**


michaelklishin:
From the looks of it, resetting those two in `close` makes sense.

FractalizeR:
Shouldn't we have a release with this bugfix?

coyl:
Guys, when a stable version comes? This bug is killing me. My project uses bundles, that use this lib and to fix it I'll have to fork these bundles and your lib =(

acasademont:
Please, we need a release for this!

michaelklishin:
I haven't gotten any response in #469, so proceeded to tag `v2.7.0` and updated the change log.

coyl:
🎉🎉🎉

vojtech-dobes:
🍹 🎆 🎺

PR #482, opened 2017-05-07
Add support for serializing DateTimeImmutable objects

IMSoP:
Rather than checking for \DateTime objects directly, check for the
\DateTimeInterface added in PHP 5.5. Since older versions are still
supported, a redundant check for \DateTime is left in place.

Reading a timestamp still always returns a DateTime object.


michaelklishin:
Thank you!

PR #491, opened 2017-06-07
DO NOT MERGE Fixed requeue param for nack and reject for more consistency.

corpsee:
Fixed requeue param for nack and reject for more consistency.

### DIFF
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -846,7 +846,7 @@ class AMQPChannel extends AbstractChannel
      * @param bool $multiple
      * @param bool $requeue
      */
-    public function basic_nack($delivery_tag, $multiple = false, $requeue = false)
+    public function basic_nack($delivery_tag, $multiple = false, $requeue = true)
     {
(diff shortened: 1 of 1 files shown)


michaelklishin:
For more consistency with what?

This is a breaking API change. Please provide a more detailed justification for it.

cc @hairyhum @nubeiro

corpsee:
This method:
```php
AMQPChannel::basic_nack($delivery_tag, $multiple = false, $requeue = false)
```

сall another method with another default `requeue` param:
```php
Protocol091::basicNack($delivery_tag = 0, $multiple = false, $requeue = true)
```
I guess default behavior for nack would be requeue = 'true' (like in `basicNack` method) because it is more useful.

michaelklishin:
I don't have an opinion on whether default value of `true` would make more sense for the `requeue` argument but this is a breaking API change in two method signatures that have been around

 * For 3 years (https://github.com/corpsee/php-amqplib/commit/919527e01e1f291c9a5f8004eec9984c46c56844)
 * or 6 years (https://github.com/corpsee/php-amqplib/commit/05eae54c17650cc94efb6e1341cca08bb387ebc1)

I'll file an issue so that we can discuss how we can make the defaults match and cause as little damage as possible. Thank you for bringing this up.

PR #500, opened 2017-07-31
Add php-enqueue/amqp-lib (queue-interop/amqp-interop) shim interface

makasim:
PR ports [eqnueue/amqp-lib](https://github.com/php-enqueue/amqp-lib) which is built on top of [ampq-interop](https://github.com/queue-interop/amqp-interop).  The solution is fully functional and already tested (I'll port tests once we agree on this). The reason to have it instead of stand alone library is that it will be adopted by more developers which is good for a amqp interop and PHP community. 

Here's [usage examples](https://github.com/php-enqueue/enqueue-dev/blob/master/docs/transport/amqp_lib.md)

The main reason of the PR is to make messaging easy to do and open new opportunities for our PHP ecosystem. I am welcome you to join the queue interop group. Let's build it together 

1. AMQP itself is a specification, not implementation and it would be good to have it reflected in PHP language. As an author of [a library](https://github.com/formapro/pvm) which relies on messaging, I do not want to choose between amqplib and amqpext. I want it to work with both! It's like `psr/log` or `psr7` but for messaging. 
2. Right now there are two main AMQP clients in PHP world: this library and [amqp ext](https://github.com/pdezwart/php-amqp) and they are not compatible. They have completely different APIs. So people build solutions that work with [the library](https://github.com/php-amqplib/RabbitMqBundle) or with [the extension](https://github.com/M6Web/AmqpBundle) only. All they start to build their own clients. Such a waste of time and resources.  
3. Right now everybody is busy building yet another AMQP client whether it's built on top of this library or not. If we are agreed on messaging standard we can do a step further and start building more advanced stuff on top of the standard. Instead of writing yet another client developers can play with [enterprise patters](http://www.enterpriseintegrationpatterns.com/). It is a lot more fun IMO. 
4. I can create a library on top of the standard that provides a delaying strategy based on dead queues or on the delay plugin without the need to the coupling on an implementation (an example of level up). Everybody can install it using composer and just use it. Right now we have loads of articles explaining how to delay messages using 100500+ different amqp libraries. 
5. I hope frameworks will be more interested in adopting MQ if they have the ability to rely on pure interfaces. 
6. The amqp-interop makes it easier to use AMQP. You learn the interfaces once and you are ready to use any compatible library. 
7. Besides, the PR is relatively small. 

I'll add tests and docs if there is an approved from maintainers

### DIFF
--- a/PhpAmqpLib/Interop/AmqpConnectionFactory.php
+++ b/PhpAmqpLib/Interop/AmqpConnectionFactory.php
@@ -1,0 +1,229 @@
+<?php
+
+namespace PhpAmqpLib\Interop;
+
+use Interop\Amqp\AmqpConnectionFactory as InteropAmqpConnectionFactory;
+use PhpAmqpLib\Connection\AbstractConnection;
+use PhpAmqpLib\Connection\AMQPLazyConnection;
+use PhpAmqpLib\Connection\AMQPLazySocketConnection;
+use PhpAmqpLib\Connection\AMQPSocketConnection;
+use PhpAmqpLib\Connection\AMQPStreamConnection;
+
+class AmqpConnectionFactory implements InteropAmqpConnectionFactory
+{
+    /**
+     * @var array
+     */
+    private $config;
+
+    /**
+     * @var AbstractConnection
+     */
+    private $connection;
+
+    /**
+     * The config could be an array, string DSN or null. In case of null it will attempt to connect to localhost with default credentials.
+     *
+     * [
+     *     'host'  => 'amqp.host The host to connect too. Note: Max 1024 characters.',
+     *     'port'  => 'amqp.port Port on the host.',
+     *     'vhost' => 'amqp.vhost The virtual host on the host. Note: Max 128 characters.',
+     *     'user' => 'amqp.user The user name to use. Note: Max 128 characters.',
+     *     'pass' => 'amqp.password Password. Note: Max 128 characters.',
+     *     'lazy' => 'the connection will be performed as later as possible, if the option set to true',
+     *     'stream' => 'stream or socket connection',
+     *     'receive_method' => 'Could be either basic_get or basic_consume',
+     * ]
+     *
+     * or
+     *
+     * amqp://user:pass@host:10000/vhost?lazy=true&socket=true
+     *
+     * @param array|string $config
+     */
+    public function __construct($config = 'amqp://')
+    {
+        if (empty($config) || 'amqp://' === $config) {
+            $config = [];
+        } elseif (is_string($config)) {
+            $config = $this->parseDsn($config);
+        } elseif (is_array($config)) {
+        } else {
+            throw new \LogicException('The config must be either an array of options, a DSN string or null');
+        }
+
+        $this->config = array_replace($this->defaultConfig(), $config);
+
+        $supportedMethods = ['basic_get', 'basic_consume'];
+        if (false == in_array($this->config['receive_method'], $supportedMethods, true)) {
+            throw new \LogicException(sprintf(
+                'Invalid "receive_method" option value "%s". It could be only "%s"',
+                $this->config['receive_method'],
+                implode('", "', $supportedMethods)
+            ));
+        }
+    }
+
+    /**
+     * @return AmqpContext
+     */
+    public function createContext()
+    {
+        return new AmqpContext($this->establishConnection(), $this->config['receive_method']);
+    }
+
+    /**
+     * @return AbstractConnection
+     */
+    private function establishConnection()
+    {
+        if (false == $this->connection) {
+            if ($this->config['stream']) {
+                if ($this->config['lazy']) {
+                    $con = new AMQPLazyConnection(
+                        $this->config['host'],
+                        $this->config['port'],
+                        $this->config['user'],
+                        $this->config['pass'],
+                        $this->config['vhost'],
+                        $this->config['insist'],
+                        $this->config['login_method'],
+                        $this->config['login_response'],
+                        $this->config['locale'],
+                        $this->config['connection_timeout'],
+                        $this->config['read_write_timeout'],
+                        null,
+                        $this->config['keepalive'],
+                        $this->config['heartbeat']
+                    );
+                } else {
+                    $con = new AMQPStreamConnection(
+                        $this->config['host'],
+                        $this->config['port'],
+                        $this->config['user'],
+                        $this->config['pass'],
+                        $this->config['vhost'],
+                        $this->config['insist'],
+                        $this->config['login_method'],
+                        $this->config['login_response'],
+                        $this->config['locale'],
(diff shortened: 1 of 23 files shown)


michaelklishin:
Can we take a step back and explain what's the goal of this PR? If `php-enqueue/amqp-lib` already exists, why it should be rolled into this client library and not continue being a separate thing?

makasim:
@michaelklishin 

The main reason of the PR is to make messaging easy to do and open new opportunities for our PHP ecosystem. I am welcome you to join the queue interop group. Let's build it together 

1. AMQP itself is a specification, not implementation and it would be good to have it reflected in PHP language. As an author of [a library](https://github.com/formapro/pvm) which relies on messaging, I do not want to choose between amqplib and amqpext. I want it to work with both! It's like `psr/log` or `psr7` but for messaging. 
2. Right there are two main AMQP clients in PHP world: this library and [amqp ext](https://github.com/pdezwart/php-amqp) and they are not compatible. So people build solutions that work with [the library](https://github.com/php-amqplib/RabbitMqBundle) or with [the extension](https://github.com/M6Web/AmqpBundle) only. 
3. Right now everybody is busy building yet another AMQP client whether it's built on top of this library or not. If we are agreed on messaging standard we can do a step further and start building more advanced stuff on top of the standard. There is no need in such a big amount of amqp clients. 
4. I can create a library on top of the standard that provides a delaying strategy based on dead queues or on the delay plugin without the need to the coupling on an implementation (an example of level up). Right now we have loads of articles explaining how to delay messages using 100500+ different amqp libraries. 
5. I hope frameworks will be more interested in adopting MQ if they have the ability to rely on pure interfaces. 
6. The amqp-interop makes it easier to use AMQP. You learn the interfaces once and you are ready to use any compatible library.

michaelklishin:
@makasim I understand the premise of shim libraries and sometimes they make sense. My question is not so much about why `php-enqueue` exists but rather why this code should reside in this library and not be a library of its own, or a part of `php-enqueue`. It's not uncommon to see both
approaches in the open source world:

 * Individual libraries provide their implementations of a popular shim library (such as logging shims)
 * Shim libraries accumulate a bunch of implementations for libraries

The former approach is common for really popular shims (SLF4J in the JVM ecosystem, for example). When a shim is not particularly ubiquitous, it provides implementations for all the underlying "implementation" libraries.

Is `php-enqueue` really so widely deployed that this client should provide an implementation and tie itself to `php-enqueue` (the shim code won't maintain itself)? It seems to be a pretty new project spearheaded by a company (or few companies).

makasim:
The solution does not rely on any enqueue lib part and does not require it. It needs only [amqp interop](https://github.com/queue-interop/amqp-interop) interfaces.

makasim:
This interop is similar to this lib https://github.com/rabbitmq/rabbitmq-jms-client and shares a lot as they both based on JMS

michaelklishin:
@makasim perhaps I should put this more bluntly (I don't mean to offend you personally r your efforts with `php-enqueue`). Why should the maintainers of this library sign up for maintaining this code, which *is* specific to `php-enqueue`? They are already quite busy. Who will maintain it if `php-enqueue` is abandoned next month because your priorities change? Why can't it be a separate library under the `php-enqueue` account?

michaelklishin:
@makasim JMS has been ubiquitous in the JVM community for over a decade. It makes sense for any messaging related project to proivde a JMS interface implementation to the extent possible, in fact, some *only* provide a JMS interface.

`php-enqueue` is not nearly as widely adopted. I suggest that any code related to `php-enqueue` incubate in that org for a couple of years and then perhaps libraries such as this one or pdezwart/php-amqp can consider maintaining their own respective shims.

Unless @nubeiro or @hairyhum feel enthusiastic about accepting this PR, I think it's not a wise move for `php-amqplib` at this time.

makasim:
I would like to stress out that the solution I am proposing has no direct relation to php-eqnueue. 

I've worked on this stuff for a year and I am going to continue, that's what I paid for. I can support.

makasim:
Symfony (one of the biggest PHP frameworks) incorporates [AMQP extension](https://github.com/symfony/symfony/pull/23315). If they decide to support AMQPlib they will have to copy past 90% of their code. If it is based on interop interfaces it remains the same. We are working on a PR that does same things as theirs but based on the interop, so this library could be used too. 

I am serious about this stuff and I am not going to leave it anytime soon.

makasim:
Also, the interop interfaces are cool, really. I am going to come back with reworked examples.

makasim:
I add `final` to classes, it would be easier to support them, fewer BC breaks and so on and we can still learn and adjust interfaces.

makasim:
Added interop examples

michaelklishin:
@makasim you are trying to sell me on the wrong idea. As I said, shim libraries can be useful. Only really popular shim libraries are supported by individual "implementations" of their interface and your projects aren't quite there yet. You are trying to make this library carry/maintain/indirectly promote your projects without offering *any* benefit for the maintainers or the majority of users (who are not aware of your interop libraries).

Your projects are objectively immature, little known, don't have any technical reasons to include their bits into libraries such as this one. Unless @nubeiro or @hairyhum really like the idea, I'd say please come back in a couple of years.

michaelklishin:
@nubeiro @hairyhum please chime in or I'm going to close this later this week.

michaelklishin:
@makasim please no more arguments about how "interop is cool, really" or I am locking this discussion. Our concern is: why should we include/maintain/support your shim libraries forever (since removing something is quite hard once you include it)? Do you have a reason that goes beyond "but wouldn't it be cool if we were all singing and dancing users of the same common API"?

makasim:
We can go on our own indeed but it'll take much more time to spread and adapt it to PHP community.  If you accept this we can go quicker and persuade other projects to join the group. That's our plan and have just started. 

Why do you need it? Well, all my arguments are based on the idea of shared\common interfaces and their wide adoption. 

I am not new to open source: I created [Payum](https://github.com/Payum/Payum) library which is quite popular (I've been supporting for 5 years) as well as other projects that I [maintained](https://github.com/liip/LiipImagineBundle) or [contributed too](https://github.com/Sylius/Sylius). 

So I am going to support this. I'll do my best so this won't bother you at all.

mvrhov:
@makasim is trying to get everyone on the same page. Like it happened with the PSR/LogInterface and PSR/CacheInterface. the newset PSR/Containerinterface. The point is that One can use the library it wants and if the library doesn't support what he needs he can then switch elsewhere without rewriting a ton of code.

michaelklishin:
@mvrhov you are beating a dead horse. See https://github.com/php-amqplib/php-amqplib/pull/500#issuecomment-319337855.

michaelklishin:
@nubeiro @hairyhum any comments, please?

hairyhum:
I don't see any change to php-amqplib in this PR. It looks like an separate library, which can perfectly exist on it's own and depend on php-amqplib.

PR #510, opened 2017-09-21
Readme update for version 2.7

ondrejfuhrer:
Hey guys, 

Since 2.7 has been released it would make sense to update the readme as well 🙂

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Contributions are **welcome** and will be fully **credited**.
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
I intentionally didn't update it in hope to wait for a few days but fair enough, we've had an RC out for 2 months or so and some were clearly using it in production.

michaelklishin:
I intentionally didn't update it in hope to wait for a few days but fair enough, we've had an RC out for 2 months or so and some were clearly using it in production.

PR #559, opened 2018-04-17
Throw exception on missed heartbeat

hairyhum:
If an IO detects missed heartbeats, it should not reconnect because
the AMQP connection and channels will not be recovered and server-side
connection and channels will be stopped on TCP connection close or
heartbeat timeout.
The current reconnection logic leads to delayed exception, if the client
waits for messages or tries to publish/ack.
It also will create a TCP connection on the server, which will consume
socket descriptors and memory.

Replace reconnect on missed heartbear with close and exception.

Since handling exceptions is a way to address connection issues in the
client, reconnection logic should be moved to exception handlers and be
higher-level, so channels and consumers can be recovered.

Addresses #309

### DIFF
--- a/PhpAmqpLib/Wire/IO/SocketIO.php
+++ b/PhpAmqpLib/Wire/IO/SocketIO.php
@@ -221,6 +221,7 @@ class SocketIO extends AbstractIO
 
     /**
      * Heartbeat logic: check connection health here
+     * @throws \PhpAmqpLib\Exception\AMQPRuntimeException
      */
     public function check_heartbeat()
     {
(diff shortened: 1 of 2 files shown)


bmichalski:
Wouldn't it break backward compatibility?

hairyhum:
@bmichalski I don't think it can be considered an incompatibility, because with `reconnect()` you would get an exception when trying to wait for a message or execute a channel method anyway (something like `Broken pipe`), because logical channels and connections are not there anymore.
This change makes it fail earlier and not pollute the server with "empty" TCP connections.

Issue #498, opened 2017-07-18
Can we specify multiple hosts while creating AMQPConnection ?


gauravgargtech:
We have set-up a HA cluster with 2 nodes and thinking to add 2 hosts addresses in RABBITMQ_SERVER_HOSTNAME, Is it possible in this library.

                self::$connection = new AMQPConnection(RABBITMQ_SERVER_HOSTNAME, RMQ_PORT, RMQ_USER_NAME, RMQ_PASSWORD);

Can we have constant "RABBITMQ_SERVER_HOSTNAME" = 'host01,host02';

and both these hosts have same Port and username/password combinations ...

michaelklishin:
As far as I can tell, it isn't. For initial connection, it should be a fairly easy feature to add (connection recovery is a different story). Feel free to look into a pull request.

PR #562, opened 2018-04-19
Introduce a method to create connection from multiple hosts.

hairyhum:
Add a new static method `create_connection` in `AbstractConnection`
to create a connection providing multiple hosts.

The method takes an array of host arrays
with host,port,username,password and vhost keys
and tries to connect to them one-by-one.
The first successful connection is returned.
If unable to connect to any of them - the last esception is thrown.

This method is a simplest way to address issues like #253 or #498
There is no bookkeeping and load balancing, hosts are
selected one-by-one starting from the first one.

Combined with connection recovery techniques it can be used
to improve reliability when using clusters.


hairyhum:
@nubeiro are you happy with us merging this?

PR #589, opened 2018-08-06
Fix socket eagain error

RackovychV:
When queue has no messages, I've received EAGAIN error code.
I do not use heartbeat in my configuration.


RackovychV:
Not fixed

PR #633, opened 2018-12-14
Fixed send_channel_method_frame

limingxinleo:


ramunasd:
IMHO this one is not for external use and should be protected. What is the issue You are trying to fix?

PR #640, opened 2019-01-22
Add some constant to AMQP exchange

dream-mo:
When we need rabbitmq exchange type, use constant replace string,such as "direct", "fanout" and so on. You can find it in class PhpAmqpLib\Exchange\AMQPExchange.


ramunasd:
In addition I would name that class as `final class AMQPExchangeType {}`  and i would strip redundant prefix from all constant names, so instead of `AMQPExhange::AMQP_EX_TYPE_DIRECT` i would prefer `AMQPExchangeType::DIRECT`.

dream-mo:
> In addition I would name that class as `final class AMQPExchangeType {}` and i would strip redundant prefix from all constant names, so instead of `AMQPExhange::AMQP_EX_TYPE_DIRECT` i would prefer `AMQPExchangeType::DIRECT`.

that great

ramunasd:
BTW, amqp extension provides same constants too.
https://github.com/pdezwart/php-amqp/blob/1205d3287df0a9ec762a6594b4fa018ed9637d21/amqp.c#L168

dream-mo:
> 谢谢你的拉动请求。你可以修改代码[`demo`](https://github.com/php-amqplib/php-amqplib/tree/master/demo)来使用这些常量吗？

Ok.I will push a new pull requests

PR #645, opened 2019-02-13
AMQPExceptionInterface should always be Throwable

derek9gag:
We found it in static code analysis. `AMQPExceptionInterface` should always be `\Throwable`, as it is intend to be caught with try-catch.


ramunasd:
You forgot support for php <7.0 :)

derek9gag:
Oh... right

PR #678, opened 2019-04-11
Fix detection of a missed server heartbeat

alexandre-abrioux:
Hi,

We had issues on our infrastructure where the network connection between our consumers and RabbitMQ would get broken from time to time and our consumers wouldn't detect it.

## Issue

After those network interruptions RabbitMQ would show that the consumers were not connected anymore. An inspection of established connections with `netstat` in our consumers containers would show the following:
```bash
root@c0b3853ca0f2:/var/www/html# netstat -tn
Active Internet connections (w/o servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State
tcp        0      0 ***.**.*.*:41088        ***.**.*.**:5672        ESTABLISHED
tcp        0      0 ***.**.*.*:41214        ***.**.*.**:5672        ESTABLISHED
```
; meaning that, according to our consumers, their `IOStream` with RabbitMQ were sill opened. They were still waiting for server incoming messages - even tho they were receiving no heartbeat as we confirmed with `tcpdump`.

## How to reproduce

I found how to reproduce the issue on my local environment with a simple docker setup.
Launch a dummy consumer connected to a RabbitMQ instance.
The consumer must be configured to run indefinitely (see below).

We are using `php-amqplib/rabbitmq-bundle` for Symfony, and the following configuration:
```yaml
old_sound_rabbit_mq:
    connections:
        default:
            host:         ...
            port:         ...
            user:         ...
            password:     ...
            vhost:        ...
            read_write_timeout:     120
            heartbeat:              60
            lazy:                   true
    consumers:
        dummy:
            connection:           default
            exchange_options:     ...
            queue_options:        ...
            auto_setup_fabric:    false
            callback:             ...
```
As you can see, we are not using the `idle_timeout` option in our consumers meaning that they wait indefinitely for new messages and aren't restarted on a regular basis - this is how we would prefer it.

Launch the consumer:

```bash
bin/console rabbitmq:consumer --env=dev -n -w dummy --debug -vvv
```

Simply disconnect the RabbitMQ container from the docker network that links it to the consumer and see what happens.
```bash
docker network disconnect [network_name] [rabbitmq_container_name]
```

### What happens

The consumer keeps running even after twice the heartbeat frame.

### What should happen

The consumer should throw an `AMQPHeartbeatMissedException`.

## How I fixed it

In case the heartbeat feature is activated, and the `$timeout` sent in the `AbstractConnection::wait_frame` method is greater that twice the heartbeat frame, I have shortened the timeout to twice the heartbeat frame (see PR). Disclaimer: I just discovered the library and there maybe is a better way to implement it.

Thank you!

### DIFF
--- a/PhpAmqpLib/Connection/AbstractConnection.php
+++ b/PhpAmqpLib/Connection/AbstractConnection.php
@@ -603,7 +603,7 @@ class AbstractConnection extends AbstractChannel
             $wait_frame_timeout = $_timeout;
             // in case heartbeats are activated prevents waiting more than 2 times the heartbeat frame
             // otherwise we would never detect a missed server heartbeat
-            if($this->heartbeat && ($wait_frame_timeout === 0 || $wait_frame_timeout > $this->heartbeat * 2)) {
+            if($this->heartbeat && !is_null($timeout) && ((double) $wait_frame_timeout === 0.0 || $wait_frame_timeout > $this->heartbeat * 2)) {
                 $wait_frame_timeout = $this->heartbeat * 2 + 1;
             }
             try {
--- a/PhpAmqpLib/Connection/AbstractConnection.php
+++ b/PhpAmqpLib/Connection/AbstractConnection.php
@@ -600,8 +600,14 @@ class AbstractConnection extends AbstractChannel
         $_timeout = $timeout;
         while (true) {
             $now = time();
+            $wait_frame_timeout = $_timeout;
+            // in case heartbeats are activated prevents waiting more than 2 times the heartbeat frame
+            // otherwise we would never detect a missed server heartbeat
+            if($this->heartbeat && ($wait_frame_timeout === 0 || $wait_frame_timeout > $this->heartbeat * 2)) {
+                $wait_frame_timeout = $this->heartbeat * 2 + 1;
+            }
             try {
-                list($frame_type, $frame_channel, $payload) = $this->wait_frame($_timeout);
+                list($frame_type, $frame_channel, $payload) = $this->wait_frame($wait_frame_timeout);
             } catch (AMQPTimeoutException $e) {
                 if ( $this->heartbeat && microtime(true) - ($this->heartbeat*2) > $this->last_frame ) {
                     $this->debug->debug_msg("missed server heartbeat (at threshold * 2)");
--- a/tests/Functional/Channel/ChannelWaitTest.php
+++ b/tests/Functional/Channel/ChannelWaitTest.php
@@ -14,8 +14,9 @@ class ChannelWaitTest extends TestCase
      * @small
      * @dataProvider provide_channels
      * @param callable $factory
+     * @param int $heartBeat
      */
-    public function should_wait_until_signal_by_default($factory)
+    public function should_wait_until_signal_by_default($factory, $heartBeat)
     {
         $this->deferSignal(0.5);
         /** @var AMQPChannel $channel */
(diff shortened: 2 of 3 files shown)


alexandre-abrioux:
#### Note
I could have used the `AbstractConnection::$read_timeout` property instead of hard-coding `$wait_frame_timeout = $this->heartbeat * 2 + 1;`.

The `$read_timeout` is already set on the stream in the `StreamIO::connect` method with the `stream_set_timeout()` function but the PHP doc says that no error is thrown is case of a timeout.

> When the stream times out, the 'timed_out' key of the array returned by stream_get_meta_data() is set to TRUE, although no error/warning is generated. 
https://www.php.net/manual/en/function.stream-set-timeout.php

https://github.com/php-amqplib/php-amqplib/blob/a0c2bef14bca396f3268c8ec89d42713e014342e/PhpAmqpLib/Wire/IO/StreamIO.php#L132-L135

alexandre-abrioux:
Hi @ramunasd and thank you for the reply!

The library was as you said stuck on an infinite `stream_select()` state in our specific case.

Sorry about the rest. I saw the `CONTRIBUTING.md` file only after opening my PR.
I then took the time to add a new test named `must_throw_exception_on_missed_heart_beat`.

I couldn't however find a way to validate the `should_wait_until_signal_by_default` test without a big rework on the library code.

For now I just accepted the fact that an `AMQPTimeoutException` is thrown when aborting the connection with a signal in case heartbeats are activated.

The issue I encountered is that when a signal is caught the IO `$result` is changed from `false` to `0`:
https://github.com/php-amqplib/php-amqplib/blob/e2f9b7c136cfc62ddcd53425385e4a66fa483e83/PhpAmqpLib/Wire/IO/AbstractIO.php#L96-L99
; and therefore - as my first commit changes the `wait_frame` timeout from `0` to two times the heartbeat frame - an `AMQPTimeoutException` exception is thrown instead of an `AMQPNoDataException`:
https://github.com/php-amqplib/php-amqplib/blob/e2f9b7c136cfc62ddcd53425385e4a66fa483e83/PhpAmqpLib/Wire/AMQPReader.php#L140-L149
It is thus not caught in the `AbstractChannel::wait` while loop:
https://github.com/php-amqplib/php-amqplib/blob/e2f9b7c136cfc62ddcd53425385e4a66fa483e83/PhpAmqpLib/Channel/AbstractChannel.php#L349-L355

I actually don't understand why the `$result` is set to `0` when a signal is caught, and not left to `false`, as we are actually losing information in this case. So I preferred coming back to you with a temporary solution first, that is accepting that the `AMQPTimeoutException` is thrown.

As a better fix we could, for instance, create a new exception class that would be thrown if heartbeats are activated and get caught in the `AbstractChannel::wait` while loop, but I'm afraid to insert some side effect by making more edits to the library.

Feel free to guide me in the way you would prefer to see it implemented. I am however not sure if I will be able to invest more time on that fix as we didn't plan to encounter it in the first place ;) I will gladly discuss more on the matter if necessary tho!

Cheers.

ramunasd:
So let's start from beginning :)

Your issue is that infinite wait() is stuck because of broken connection and no exception is thrown.
Expected result should be raised AMQPConnectionClosedException, not AMQPHeartbeatMissedException, because latter indicates other type of problems - when consumer spends too much time on task and library has no chance to send heartbeat message.

So how to detect broken connection? We could constantly send heartbeat message and throw exception if delivery failed. It means, long stream_select() timeouts should be split into smaller parts, with heartbeat checks in between. This is what I mean with `better passed time calculations` :) The first part was to separate logical actions and implementation(AbstractIO->select() vs. AbstractIO->do_select()), which was done in previous version.

Now logic is in one place and we can implement time splitting. Inside of `AbstractIO->select()`, instead of calling `$this->do_select()` once, we could call it many times, with shorter and not too big timeouts. Those timeouts must be calculated based on last read/write time and not exceed next heartbeat deadline, obviously. Some possible scenarios here:
- timeout !=0, shorter than next heartbeat - select once and return
- timeout !=0, longer than next neartbeat - select and check heartbeat many times
- timeout =0, selecy and check heartbeat infinitely

What do You think @alexandre-abrioux ?

ramunasd:
stale & conflicts

PR #695, opened 2019-05-27
rm ext-bcmath dependency

terrafrost:
See #694

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,12 +12,6 @@ It's been tested against [RabbitMQ](http://www.rabbitmq.com/).
 
 **Requirements: PHP 5.3** due to the use of `namespaces`.
 
-**Requirements: bcmath extension** This library utilizes the bcmath PHP extension. The installation steps vary per PHP version and the underlying OS. The following example shows how to add to an existing PHP installation on Ubuntu 15.10:
-
-```bash
-sudo apt-get install php7.0-bcmath
-```
-
 The library was used for the PHP examples of [RabbitMQ in Action](http://manning.com/videla/) and the [official RabbitMQ tutorials](http://www.rabbitmq.com/tutorials/tutorial-one-php.html).
 
 Please note that this project is released with a [Contributor Code of Conduct](CODE_OF_CONDUCT.md). By participating in this project you agree to abide by its terms.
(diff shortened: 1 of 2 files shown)


ramunasd:
Should we not add bcmath extension as suggested package with explaination why it is needed?

ramunasd:
@terrafrost This one looks like an easy way to get rid of bcmatch extension. But the main problem is not extension itself. It is complicated and messy way in AMQPReader/AMQPWriter how to deal with big integers in 32/64 bit systems.
Take a look at \PhpAmqpLib\Wire\AMQPReader::read_longlong()
Could we use phpseclib directly here without all those checks and workarounds? It would be really nice improvement in terms of performance and simplicity.

terrafrost:
> Take a look at \PhpAmqpLib\Wire\AMQPReader::read_longlong()
Could we use phpseclib directly here without all those checks and workarounds? It would be really nice improvement in terms of performance and simplicity.

Reviewing `read_longlong`...  yah - phpseclib\Math\BigInteger could definitely be used, directly, there. phpseclib uses BigInteger with 64-bit precision here:

https://github.com/phpseclib/phpseclib/blob/e2256f42673633e0e5ff1335faf3a3aa059a894f/phpseclib/Crypt/Hash.php#L891

(PHP didn't introduce support for sha512/224 or sha512/256 until PHP 7.1.0 and phpseclib 2.0 works on PHP 5.3+)

That said, I think `read_php_int` and `read_long` could be updated to work better on 32-bit systems:

```php
list(, $res) = unpack('N', "\xFF\xFF\xFF\xFF");
if ($res < 0) {
    $res = 0x80000000 + ($res & 0x7FFFFFFF);
}
```

Altho PHP doesn't always have 64-bit int support it always has support for [binary64](https://en.wikipedia.org/wiki/Double-precision_floating-point_format), which has a significand precision of 53 bits.

I'll try to do a PR as time permits! The three-day weekend is coming to an end lol.

That said, bcmath_compat would still be useful for stuff like this:

https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Wire/AMQPDecimal.php

ramunasd:
Cool, many thanks @terrafrost

terrafrost:
Superseded by #737

PR #744, opened 2019-10-03
Allow cluster support for AMQPLazyConnection by re-initializing StreamIO on node failure

FabriZZio:
This PR aims to allow AMQPLazyConnection to support RabbitMQ clusters (multiple hosts).

Since AMQPLazyConnection does not connect on construct, support for multiple hosts was not available here. By catching failures on (actual) connect, we try to re-initialize IO, allowing to connect to other nodes.

Important: since `user`, `pass` and `vhost` are stored on the `AbstractConnection` object and we are only re-initializing IO, this will only work if different nodes have the same credentials. (which is most likely the case)

In order to run the included extra tests successfully, the rabbitmq node defined in docker-compose needs to be running.

### DIFF
--- a/PhpAmqpLib/Connection/AMQPLazyConnection.php
+++ b/PhpAmqpLib/Connection/AMQPLazyConnection.php
@@ -69,6 +69,16 @@ class AMQPLazyConnection extends AMQPStreamConnection
      */
     protected function connect()
     {
+        if ($this->isConnected()) {
+            return;
+        }
+
+        // only when using AbstractConnection::create_connection for cluster support, allow multiple hosts
+        if (!self::$hosts)
+        {
+            return parent::connect();
+        }
+
         $latest_exception = null;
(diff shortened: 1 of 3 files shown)


ramunasd:
Hi @FabriZZio, thanks for PR. I really appreciate it ;)
However, this one looks like a hacky workaround and adds some unneeded complexity to the already complicated library. This would make any future refactoring harder. I believe you can implement a custom wrapper around this library for lazy multi host support.
So factory pattern over exploding logic?

FabriZZio:
@ramunasd Thanks for looking into the PR. It is very hacky indeed. To me, it seems like the place where the multi-host fallback logic exists, is in the wrong place. It should be available closer to the actual 'connect'  instead of the initialization of an `AbstractConnection`. (for all types of connections)

We'll implement this in a separate wrapper in our application.

Thanks for your support!

PR #753, opened 2019-10-30
Fix AMQPTimeoutException handling

kozlice:
See issue #752


jenssegers:
@lukebakken Will there be a release with this fix soon?

PR #775, opened 2020-01-23
Async heartbeats

igrizzli:
If time of processing message exceed heartbeat time, consumer will fall with exception.
With PR adds sending heartbeats to server during message processing.

Based on https://wiki.php.net/rfc/async_signals which was added in PHP 7.1

### DIFF
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -1040,7 +1040,9 @@ class AMQPChannel extends AbstractChannel
         );
 
         if (isset($this->callbacks[$consumer_tag])) {
-            call_user_func($this->callbacks[$consumer_tag], $message);
+            $this->connection->executeWithAsyncHeartbeat(function () use ($consumer_tag, $message) {
+                call_user_func($this->callbacks[$consumer_tag], $message);
+            });
         }
     }
 
--- a/PhpAmqpLib/Connection/AbstractConnection.php
+++ b/PhpAmqpLib/Connection/AbstractConnection.php
@@ -938,6 +938,21 @@ class AbstractConnection extends AbstractChannel
     }
 
     /**
+     * Execute callback with async heartbeats
+     *
+     * @param callable $callback
+     * @return mixed
+     */
+    public function executeWithAsyncHeartbeat(callable $callback)
+    {
+        $this->io->enableAsyncHeartbeat();
+        $result = $callback();
+        $this->io->disableAsyncHeartbeat();
+
+        return $result;
+    }
+
+    /**
      * Check connection heartbeat if enabled.
      * @throws AMQPHeartbeatMissedException If too much time passed since last connection activity.
      * @throws AMQPConnectionClosedException If connection was closed due to network issues or timeouts.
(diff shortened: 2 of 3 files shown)


ramunasd:
Thanks @igrizzli , it's nice idea. But i don't like the way it was implemented. Can you leave IO classes as it is and make this feature as optional? Code complexity is still high and we dont want to increase it even more.

igrizzli:
> Thanks @igrizzli , it's nice idea. But i don't like the way it was implemented. Can you leave IO classes as it is and make this feature as optional? Code complexity is still high and we dont want to increase it even more.

I can move it to connection level, but it need to modify IO anyway:
- don't check server heartbeats while async enabled
- Modify $last_read property

I choosed IO level to implement this, because heartbeat logic implemented there. If you have another vision of place for this code, can you explicitly say there i should place it?

About optional feature, i see two ways to make it:
1. Use ENV AMQP_ASYNC_HEARTBEATS
2. Turn it on only if pcntl_async_signals already enabled.

Which way is better  by your opinion?

ramunasd:
I think this handler should be implemented as separate class, maybe static one. And everyone could optionally "register" this handler, if possible ofcourse. THere is no need to change anything within connection or IO layer, cause you can use this method https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Connection/AbstractConnection.php#L948

igrizzli:
>THere is no need to change anything within connection or IO layer, cause you can use this method https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Connection/AbstractConnection.php#L948

I can't use this method because i should:
- don't check server heartbeats while async enabled, but send from client
- Modify $last_read property, in the end of async heartbeat

It also not control server heartbeats, so it has sense to be on only during handling message, so connection should be also modified.

So even if logic of alarms will be in separated class, with static - on - off, i can't avoid modification of IO and Connection level.

I tried to implement it in RabbitMqBundle first, but have to fallback to php-amqplib because it will not work without modification io level.

igrizzli:
We also can split check_heartbeat to check_server_heartbeat and send_heartbeat, but it will BC.

ramunasd:
Please elaborate more why you can't check server heartbeats while async is enabled?

igrizzli:
Now implementation of check server heartbeats check last time when we read something from socket. 
If you processing message you read nothing from socket, until you finish.

ramunasd:
You are right, we cannot read from connection and ignore waiting frames.

Still this isn't reason to increase code complexity so much.

Anyways, I cannot accept this PR without any tests and demo. See https://github.com/php-amqplib/php-amqplib/blob/master/CONTRIBUTING.md

ramunasd:
@igrizzli 
In addition it would be very nice to know how async signals works in those situations:
* consumer executes long&blocking database(let's say mysql) queries
* consumer waits for network response, let's say http request
* consumer executes long running system process
* consumer just sleeps, `sleep(60)`

When and how execution jums over different parts of code? Is this same stack trace? 

If You don't have time for this then just let me know. I will continue on this on my own.

igrizzli:
@ramunasd 
I plan to refactor this and add tests on next week.

> * consumer executes long&blocking database(let's say mysql) queries
> * consumer waits for network response, let's say http request
> * consumer executes long running system process
> * consumer just sleeps, `sleep(60)`

For listed cases signals will be handled after blocking operation will be finished.

> When and how execution jums over different parts of code? Is this same stack trace?
It use internal engine hook
http://www.phpinternalsbook.com/php7/extensions_design/hooks.html#overwrite-interrupt-handler

ramunasd:
Closing in favor of #815

Issue #468, opened 2017-02-19
PHP 7.1 has asynchronous signal handling


dzuelke:
See https://wiki.php.net/rfc/async_signals

Basically, in 7.1+, the `pcntl_signal_dispatch()` call in the loop is no longer necessary.

ramunasd:
I had this idea in my head for a long time. At the first glance it looks like opportunity to get rid of some code and rely purely on PHP internals. But there is two gotchas:
- what if async signals are disabled? and they are by default. In this case signal handling could work only in asynchronous manner and only if explicitly enabled.
- it's always better to dispatch signals only in specific time moments, before or after callback, but not in the middle of execution. Then we can be sure that signal handler wont break any internals.

This is the reasons why I want to leave signal handling as it is now. Eveyrone can choose what works best, either synchronous or asynchronous signals. More options is always better.

PR #776, opened 2020-01-23
Handle exit signals async if possible

igrizzli:
Implement #468

### DIFF
--- a/PhpAmqpLib/Wire/IO/AbstractIO.php
+++ b/PhpAmqpLib/Wire/IO/AbstractIO.php
@@ -89,7 +89,7 @@ abstract class AbstractIO
             throw new AMQPIOWaitException($e->getMessage(), $e->getCode(), $e);
         }
 
-        if ($this->canDispatchPcntlSignal) {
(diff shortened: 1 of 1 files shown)


ramunasd:
Sorry, this does not make sense to me. If You enable async signals in your consumer, then just define `AMQP_WITHOUT_SIGNALS` as `true` and it wont dispatch anything.

igrizzli:
> Sorry, this does not make sense to me. If You enable async signals in your consumer, then just define `AMQP_WITHOUT_SIGNALS` as `true` and it wont dispatch anything.

Removing of this code in 3.x.x milestone, as i understand, this way better than removing, so if it's not let's decline this PR.

ramunasd:
There is no plans to remove `pcntl_signal_dispatch()`. Very likely 3.x.x will support PHP 7.0.

igrizzli:
> There is no plans to remove `pcntl_signal_dispatch()`. Very likely 3.x.x will support PHP 7.0.

it's marked as 3.x.x milestone here:
https://github.com/php-amqplib/php-amqplib/issues/468

PR #777, opened 2020-01-31
Adding an SSL example with comments

jrespeto:
Adding an SSL example with comments


ramunasd:
Thanks @jrespeto can You explain difference between this and old `demo/ssl_connection.php` ?
Can You make it at least PSR-2 compliant?

ramunasd:
4 months no response, closing.

PR #778, opened 2020-02-05
fix: ensure hosts is an array, otherwise latest_exception can be null

mr-feek:
I ran into a scenario where I was accidentally passing an empty array of hosts to `AbstractConnection::create_connection`. 

In this case, the error presented to me was `Error: Can only throw objects`, because `$latest_exception` was `null`.


ramunasd:
Thanks for contributing @mr-feek!
Can You remove default value for `$hosts`? When methods have some defaults, then everyone might think that is not neccessary to provide any value. Here `$hosts` is mandatory parameter.
Also, please add docblock for this method.
Thanks

mr-feek:
How's that look @ramunasd ?

PR #817, opened 2020-08-11
AbstractIO::select() never returns false

szepeviktor:
Discovered by @phpstan

https://github.com/php-amqplib/php-amqplib/blob/c04881d9936cc49a7acd1041f8a4926b40e3265e/PhpAmqpLib/Wire/IO/AbstractIO.php#L96-L99


szepeviktor:
Please consider running `phpstan analyze PhpAmqpLib/ -l 4` in CI.
You may consider higher levels!

ramunasd:
Probably You should remove `@throws` annotation as well? @szepeviktor

szepeviktor:
@ramunasd Done.

PR #818, opened 2020-08-24
Update AbstractIO.php

SloEddy:
When running long-lasting tasks on a queue with heartbeat enabled, if you temporarily disable the heartbeat checks for the duration of the job being handled and then reenable heartbeats immediately after it would never be able to recover without manually writing a heartbeat to the stream. If we just send a heartbeat if one is necessary immediately after reenabling the heartbeat checks it should allow the following use case to work;

handling process consuming a job -> receives a job that is expected to last a long time -> disables heartbeats -> handles job -> reenables heartbeat -> continues listening for futher jobs

### DIFF
--- a/PhpAmqpLib/Wire/IO/AbstractIO.php
+++ b/PhpAmqpLib/Wire/IO/AbstractIO.php
@@ -180,7 +180,13 @@ abstract class AbstractIO
     public function reenableHeartbeat()
     {
         $this->heartbeat = $this->initial_heartbeat;
-
+        
+        // check if it is time for client to send a heartbeat
+        $now = microtime(true);
+        if (($this->heartbeat / 2) < $now - $this->last_write) {
+            $this->write_heartbeat();
+        }
+        
         return $this;
     }
 
--- a/tests/Functional/Connection/ConnectionClosedTest.php
+++ b/tests/Functional/Connection/ConnectionClosedTest.php
@@ -188,9 +188,9 @@ class ConnectionClosedTest extends AbstractConnectionTest
     public function must_not_throw_exception_missed_heartbeat_disabled($type, $size)
     {
         $channel = $this->channel_create($type, [
-            'keepalive' => false,
+            'keepalive' => true,
             'heartbeat' => $heartbeat = 1,
-            'timeout' => 3,
+            'timeout' => 10,
         ]);
         $io = $channel->getConnection()->getIO();
 
--- a/tests/Functional/Connection/ConnectionClosedTest.php
+++ b/tests/Functional/Connection/ConnectionClosedTest.php
@@ -189,8 +189,8 @@ class ConnectionClosedTest extends AbstractConnectionTest
     {
         $channel = $this->channel_create($type, [
             'keepalive' => true,
-            'heartbeat' => $heartbeat = 1,
-            'timeout' => 10,
+            'heartbeat' => $heartbeat = 15,
+            'timeout' => 60,
         ]);
         $io = $channel->getConnection()->getIO();
 
--- a/tests/Functional/Connection/ConnectionClosedTest.php
+++ b/tests/Functional/Connection/ConnectionClosedTest.php
@@ -171,6 +171,56 @@ class ConnectionClosedTest extends AbstractConnectionTest
     }
 
     /**
+     * Try to write(publish) to connection after heartbeat was temporarily disabled
+     * @test
+     * @medium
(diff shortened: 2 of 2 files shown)


rdcx (approved):

ramunasd:
Thanks @SloEddy for Your contribution. How and why heartbeats are disabled during long jobs?
BTW, tests are mandatory, please check https://github.com/php-amqplib/php-amqplib/blob/master/CONTRIBUTING.md

SloEddy:
@ramunasd apologies for missing test. Obviously a first time for me and I am not very familiar with the test structure here. I pushed a test that, hopefully, shows what I am trying to fix.

Essentially AMQPHeartbeatMissedException is thrown regardless if the connection is still active or not, if the heartbeat is intentionally disabled and a long-running synchronous task is executed, then after reenabling the heartbeat the exception is thrown immediately at first interaction with the wire. This is due to the checkBrokerHeartbeat function being fired before the write_heartbeat after disabling it directly.

I will clean up the branch once I figure out how to make this test do what I want it to do!

SloEddy:
Also to answer your question more clearly, I have a use case where a long-running job is running in a process that both consumes and publishes messages. It must always acknowledge a message after it has been handled, now because I would like to keep heartbeat turned on at most times, I need a mechanism to disable the heartbeat when that long-lasting job will execute. Here's an sample of the code I use and how I worked around the 'problem' at the application level. Without that heartbeat write section the exception is thrown and an ack is never sent, which is clearly not desirable.

```
// Get the IO stream reference
$io = $resolver->getConnection()->getIO();
// Disable heartbeat checks on the stream
$io->disableHeartbeat();

// Handle the job
$result = $this->handleMessage($message);

// Prepare the heartbeat packet
$pkt = new AMQPWriter;
$pkt->write_octet(8);
$pkt->write_short(0);
$pkt->write_long(0);
$pkt->write_octet(0xCE);
// Write a heartbeat to the stream
$io->write($pkt->getvalue());     
// Reenable heartbeat checks on the stream           
$io->reenableHeartbeat();                
// Let it verify the heartbeat is still valid to ensure we are still online
$io->check_heartbeat();
// Acknowledge the message
if ($result) {
    $resolver->acknowledge($message);
} else {
    $resolver->reject($message);
}
```

ramunasd:
Actually `AbstractIO::check_heartbeat()` checks timings and sends heartbeat. No need to do it manually. Just call `AbstractConnection::checkHeartBeat()` as many times as You can and everything will be ok.
Personally I still don't understand why You disable heartbeats and send it manually. Doesn't makes sense at all.
Also, please squash commits with relevant message once this PR is ready.

SloEddy:
`AbstractConnection::checkHeartBeat()` just calls `AbstractIO::check_heartbeat()` directly. If you call this after heartbeat has been missed, it will throw AMQPHeartbeatMissedException without trying to send a heartbeat. That is because `AbstractIO::checkBrokerHeartbeat()` gets called before a heartbeat is written, thus throwing the exception.

IF we set heartbeat to 0 for a long time, then once you reenable the heartbeat the current logic will always throw a AMQPHeartbeatMissedException, the manual sending of a heartbeat is made so that the last_write becomes recent and allows the `AbstractIO::checkBrokerHeartbeat()` to pass without throwing an exception.

Maybe I am doing something wrong but if you look at the test maybe that'll explain my point a bit better. I believe the following logic should be supported to allow slow synchronous tasks to complete without triggering an exception.

```
// Disable heartbeat
$io->disableHeartbeat();

// miss heartbeat
sleep($heartbeat * 2 + 1);

// Reenable heartbeat
$io->reenableHeartbeat();
```

SloEddy:
Right didn't plan on closing inadvertently but will open a cleaner one, one where tests hopefully pass.

ramunasd:
`AbstractIO::checkBrokerHeartbeat()` throws exception only if there was no activity in last `$heartbeat * 2 + 1`seconds. After this period of time broker consider connection as dead and close it[1]. Unacknowledged messages are being dispatched to other consumers.

Do You say that this check is incorrect and we should allow writes despite reached timeouts?

[1]

SloEddy:
Nope @ramunasd you are correct, I was reading that heartbeat wrong - the unit test in the #820 now is telling me the whole story. Apologies I didn't quite get the problem here - I thought the heartbeat * 2 +1 is library side rather than broker-side. I see now that the only correct route in this case is a reconnect.

PR #820, opened 2020-08-25
Heartbeat disable and reenable after missed heartbeat no longer throws.

SloEddy:
I inadvertently closed this PR https://github.com/php-amqplib/php-amqplib/pull/818 so reopening as new.

@ramunasd tagging I cannot reopen the old PR so I am here now. Hopefully this one is clean.


SloEddy:
As per conversation in #818 this was a misunderstanding on my part, when heartbeat * 2 + 1 is reached the broker terminates the connection so we should in fact be throwing an exception rather than attempt to write anyway.

PR #836, opened 2020-10-06
perform CI tests using github actions

ramunasd:
Use github actions to run PHPUnit tests on each push & PR. Pretty much like travis-ci.
Only linux host for a while. Will add more platforms later for better coverage.


lukebakken (approved):
Great! Thank you.

Issue #844, opened 2020-10-29
AMQPProtocolException phpdoc arguments type annotations are swapped


zerkms:
**Version(s) affected**: 2.12.1

**Description**  

At the moment it's

```php
    /**
     * @param string $reply_code
     * @param int $reply_text
     * @param array $method_sig
     */
```

should be

```php
    /**
     * @param int $reply_code
     * @param string $reply_text
     * @param array $method_sig
     */
```

**How to reproduce**  

**Possible Solution**  

**Additional context**

PR #845, opened 2020-10-29
Fixed AMQPProtocolException phpdoc arguments type annotations

zerkms:
Fixes https://github.com/php-amqplib/php-amqplib/issues/844

### DIFF
--- a/PhpAmqpLib/Exception/AMQPException.php
+++ b/PhpAmqpLib/Exception/AMQPException.php
@@ -19,8 +19,8 @@ class AMQPException extends \Exception
     public $args;
 
     /**
-     * @param string $reply_code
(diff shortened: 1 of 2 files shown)


ramunasd:
Thanks for doc improvement. Could You fix other doc blocks as well?

zerkms:
I only happened to find this one (because I needed to create an instance of `AMQPProtocolException` exception).

I will have a look into it on monday then.

zerkms:
It looks like only `AMQPException` is also affected?

zerkms:
I checked the other class (`AMQPException`) as well.

Ideally this (and basically every other) project needs to have psalm to ensure it does not happen.

PR #852, opened 2020-12-12
Test against php 8.0

sergiy-petrov:


sergiy-petrov:
Seems it's not possible to support 5.6 and 8.0 at the same time. Wondering should we remove support of old versions of php (5.6, 7.0, 7.1 and 7.2) ?

ramunasd:
Thanks and yes, we will remove php5.6 support first. Very likelly it will happen in next major version 3.0. Next year at best.

PR #854, opened 2020-12-15
Fix Required parameter before optional parameter

gpupo:
Fix warning Required parameter ``$io`` follows optional parameter ``$vhost`` showed at **PHP 8**

### DIFF
--- a/PhpAmqpLib/Connection/AbstractConnection.php
+++ b/PhpAmqpLib/Connection/AbstractConnection.php
@@ -180,7 +180,7 @@ abstract class AbstractConnection extends AbstractChannel
         $login_method = 'AMQPLAIN',
         $login_response = null,
         $locale = 'en_US',
-        AbstractIO $io,
+        AbstractIO $io = null,
         $heartbeat = 0,
         $connection_timeout = 0,
         $channel_rpc_timeout = 0.0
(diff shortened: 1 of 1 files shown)


ramunasd:
Thanks for yet another attempt to silence those anoying deprecation warnings :)
Unfortunatelly, I cannot accept this, because docblock is not valid anymore and there is no check in code whether $io has correct value.

gpupo:
> Unfortunatelly, I cannot accept this, because docblock is not valid anymore and there is no check in code whether $io has correct value.

@ramunasd another way to fix is remove the optionals values before ``$io``since you always need to specify this parameters to achieve $io 

Can you accept now ?

ramunasd:
Now it looks much better. Can You confirm that this does not break anything?

gpupo:
@ramunasd nothing break

ramunasd:
Any change in any function arguments, including constructor, automatically is considered as BC. Let's assume somebody wants to extend this class directly and relies on default values. So it breaks compatibility. And it should be done not in patch version like 2.12.2

gpupo:
Agree. Put changes like that in a major release.

ramunasd:
We will completelly refactor constructors in next major version. This PR won't help in any way.

PR #855, opened 2020-12-23
Fix php 8 required parameters after optional

YerlenZhubangaliyev:


ramunasd:
Hello, You made a lot of arguments as optional with default value as empty string. This is serious breaking change(BC). Can You honestly confirm that this wont brake anything?

Hovewer, I cannot accept this.

PR #857, opened 2020-12-24
phpseclib3

kaizer666:
use phpseclib version 3.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build/
 docs/
 tools/
 phpDocumentor.phar
+.idea
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "php": ">=5.6.3",
         "ext-sockets": "*",
         "ext-mbstring": "*",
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ build/
 docs/
 tools/
 phpDocumentor.phar
+.idea
```
ramunasd, line 8: Please use Your own global git config for those ignores.

composer.json
```diff
@@ -32,7 +32,7 @@
         "php": ">=5.6.3",
         "ext-sockets": "*",
         "ext-mbstring": "*",
-        "phpseclib/phpseclib": "^2.0.0"
+        "phpseclib/phpseclib": "*"
```
ramunasd, line 35: Asterix allows any version, even lower than 2.0. This can lead to unexpected runtime exceptions.

ramunasd (changes requested):

ramunasd:
2 week no response and logically wrong. Closing.

PR #859, opened 2020-12-28
feat (Code Style) start integration PSR-2.

Yozhef:
I think we need to take small steps so little by little I will implement changes =)
```
Now exclude:
        <exclude name="PSR1.Methods.CamelCapsMethodName.NotCamelCaps"/>
        <exclude name="PEAR.Functions.ValidDefaultValue"/>
```

What new:
 - add code style check in GitHub Action
 - update squizlabs/php_codesniffer -> 3.5 version
 - update code
 - update gitignore file


ramunasd (approved):

.gitattributes
```diff
@@ -19,3 +19,8 @@
 /phpunit.xml.dist export-ignore
 /spec export-ignore
 /tests export-ignore
+
+###> squizlabs/php_codesniffer ###
```
ramunasd, line 23: Those should be in `.gitignore` ?

ramunasd (changes requested):

.gitattributes
Yozhef: Yes.There is no need to comment on these files, as they will only lead to more confusion. In other open projects, they are also ignored =)

ramunasd (approved):

ramunasd:
Please squash and rebase commits.

PR #861, opened 2020-12-30
Start Work PSR-12

Yozhef:
This PR start work update library to PSR-12


Yozhef:
Good night @ramunasd,  help to determine the development of this PR - these changes will require extensive testing but would like to move away from the snap case towards camelCase.

ramunasd:
Thanks for Your work, but we won't change public API as it completely breaks compatibility between versions and introduces too much work for all dependants. This is not planned and gives no real benefits. I mean, code structure and design problems will stay same.

Can You submit other PR without public API changes? Let's do it in small but confident steps ;)

Yozhef:
@ramunasd 
Thanks for the feedback.
If I understood you correctly, do you mean not to change the Channel?

If to divide into small parts - I would begin with change of style of parameters of functions and classes - as in PHP 8.0 -> it will be already possible to be bound to names of parameters

Please tell me more details that will change before the meeting so I can help. We can ignore these rules and just come to PSR-12 without them.

.gitattributes
```diff
@@ -19,3 +19,8 @@
 /phpunit.xml.dist export-ignore
 /spec export-ignore
 /tests export-ignore
+
+###> squizlabs/php_codesniffer ###
+/.phpcs-cache
```
andrew-demb, line 24: Is it should be in .gitignore, not .gitattributes?

ramunasd:
> If I understood you correctly, do you mean not to change the Channel?

By saying "public API" I mean all public properties and methods, in some cases even protected properties and methods(if not marked as final). Everything that can cause inheritance problems in exernal calls or inheritance. According semantic versioning[1] all changes in minor and patch versions must be completely backward compatible, without any exceptions.

Also, I would like to suggest split these changes in even smaller parts, by functional scope. Let's say - tests, helpers, exceptions, connections, channel. There must be ability to check code style locally, can You add composer shortcuts?

[1] https://semver.org/

ramunasd:
Please squash and rebase commits.

ramunasd:
outdated and breaking changes

PR #863, opened 2021-01-13
Fix for PHP 8

aimeos:


ramunasd:
Please check previous PR regarding support for PHP8.

Issue #865, opened 2021-01-17
PCNTLHeartbeatSender would be never triggered again when connection in writing status 


MaxwellZY:
```php
private function registerListener($interval)
{
    pcntl_signal(SIGALRM, function () use ($interval) {
        if (!$this->connection || $this->connection->isWriting()) {
            return;
        }

        if (!$this->connection->isConnected()) {
            $this->unregister();
            return;
        }

        if (time() > ($this->connection->getLastActivity() + $interval)) {
            $this->connection->checkHeartBeat();
        }

        pcntl_alarm($interval);
    });
}
```
When connection is writing, should trigger the checkHeartBeat later but not just return, the following code is fixed

```php
private function registerListener($interval)
{
    pcntl_signal(SIGALRM, function () use ($interval) {
        if (!$this->connection) {
            return;
        }

        if (!$this->connection->isConnected()) {
            $this->unregister();
            return;
        }
        
        if ($this->connection->isWriting()) {
            pcntl_alarm($interval);
            return;
        }

        if (time() > ($this->connection->getLastActivity() + $interval)) {
            $this->connection->checkHeartBeat();
        }

        pcntl_alarm($interval);
    });
}
```

ramunasd:
@laurynasgadl Could You check this? I bet it's quite easy to add another functional test and fix it.

laurynasgadl:
@ramunasd sure, I'll look into it

ramunasd:
will be fixed in 2.12.2

PR #866, opened 2021-01-18
fix PCNTL heartbeat signal registration

laurynasgadl:
This is a fix for PCNTL heartbeat signal registration/chaining issue described in https://github.com/php-amqplib/php-amqplib/issues/865


ramunasd:
Many thanks 👍

PR #879, opened 2021-02-13
Fix: Required parameter $io follows optional parameter $vhost

olvlvl:
Hi everyone,

This PR fixes an issue with PHP 8.0: "Required parameter $io follows optional parameter $vhost".

Because `$io` is now nullable I'm throwing an exception if it happens, to have similar behavior as before.

I've tested the branch with my service and it works as expected.

An alternative would be to move the `$io` parameter before optional ones because classes extending the abstract don't have that in their constructor but that would be a breaking change for external projects.

Best regards,

Olivier


andrew-demb:
See https://github.com/php-amqplib/php-amqplib/pull/858

olvlvl:
Thanks for the link @andrew-demb. Looking forward to the release.

PR #880, opened 2021-02-15
feat(CodeCov) add Codecov phpunit code coverage.

Yozhef:


.github/workflows/phpunit.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         php-version: ${{ matrix.php }}
         tools: composer:v1
         extensions: bcmath, curl, dom, mbstring, pcntl, sockets, xml
-        coverage: none
+        coverage: xdebug
```
ramunasd, line 51: It will slow down all preparation and tests, yes?

.github/workflows/phpunit.yml
```diff
@@ -58,5 +58,10 @@ jobs:
     - name: Wait for broker service
       run: php ./tests/wait_broker.php
     - name: PHPUnit tests
-      run: ./vendor/bin/phpunit --exclude-group proxy
+      run: ./vendor/bin/phpunit --exclude-group proxy --coverage-clover=coverage.xml
```
ramunasd, line 61: Are You sure we need to calculate coverage for all PHP versions?

ramunasd (changes requested):

.github/workflows/phpunit.yml
Yozhef: This is a good question - probably only on the latest version of php? For example now php === 7.4?

.github/workflows/phpunit.yml
Yozhef: Maybe move this to matrix -> if this value === xdebug -> add  xdebug  and  run coverage ?

.github/workflows/phpunit.yml
ramunasd: How does codeconv works? Do we need registered something, or push those coverage files to somewhere? I've never used it, so any introduction would be helpful ;)

Yozhef: You need to start a project at https://app.codecov.io
Then you have a project account example this - https://codecov.io/gh/symfony-bundles/redis-bundle/pull/31?src=pr&el=h1
All pr -> upload info in the system then you see graphic and codcov system push in pr statistic https://github.com/symfony-bundles/redis-bundle/pull/31

ramunasd: Thanks. You should know, that neither me or Luke can install/setup new integrations. We both are maintainers, not owners. And we cannot get more permissions as current owners are not responding to our requests.
ATM it's not possible to setup codeconv, sorry.

nubeiro: Hi, hopefully I did the right thing. Could someone confirm it works as expected?

nubeiro: Also, is there anything I can do to help @ramunasd with permission issues?

Yozhef: https://app.codecov.io/gh/php-amqplib/php-amqplib/commits?page=1

ramunasd: @nubeiro what a surprise! and actually yes, You can help me. It would be nice to have some permissions for https://github.com/php-amqplib/RabbitMqBundle as it is crying for some maintenance. Also, would be nice to become member of php-amqplib organization.

ramunasd (approved):

.github/workflows/phpunit.yml
ramunasd: @nubeiro and appveyor integration as well, currently it runs on my personal account

PR #893, opened 2021-02-24
Allow to use SSL connection as lazy

adombrovsky:


PhpAmqpLib/Connection/AMQPLazySSLConnection.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace PhpAmqpLib\Connection;
+
+class AMQPLazySSLConnection extends AMQPSSLConnection
+{
+     /**
+     * Gets socket from current connection
+     *
+     * @deprecated
```
ramunasd, line 10: This probably will go to 3.0, so let's drop deprecated methods.

ramunasd (changes requested):

PhpAmqpLib/Connection/AMQPLazySSLConnection.php
adombrovsky: done

adombrovsky: @ramunasd can we also port this change for the 2.x version? I would like to use it there as well.

ramunasd (approved):

PR #900, opened 2021-03-01
added AMQPLazySSLConnection for 2.x version

adombrovsky:
This is a port of https://github.com/php-amqplib/php-amqplib/pull/893 PR in order to have this functionality for 2.x version


ramunasd:
We won't add new features to 2.x series. Only very important bug fixes.

PR #912, opened 2021-04-24
Security issue in phpseclib package

sergiy-petrov:
Updating phpseclib package minimum versions. Please find the link below for more details 

https://github.com/advisories/GHSA-vf4w-fg7r-5v94


ramunasd:
Thanks for pointing it out. Actually it is already know as dependabot alert is enabled and we got the message.
The thing is that we neither use affected code nor are responsible for security upgrades.

PR #916, opened 2021-05-10
Fixed composer php version constraint

rkrx:
Can't install using php7.4.
See here: https://jubianchi.github.io/semver-check/#/^7.0|~8.0.0/7.4
Fix: https://jubianchi.github.io/semver-check/#/^7.0||~8.0.0/7.4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         }
     ],
     "require": {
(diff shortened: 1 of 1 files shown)


ramunasd:
Thanks!

PR #917, opened 2021-05-14
Improved examples and dosc

corpsee:
Removed using of internal property `AMQPMessage::delivery_info` from docs and examples.


ramunasd:
Thanks!

PR #923, opened 2021-06-19
Use correct default for read_write_timeout in AMQPStreamConnection try_create_connection

bezhermoso:
Looks like this was missed in this revert: https://github.com/php-amqplib/php-amqplib/pull/726


bezhermoso:
Actually, this looks intentional...

PR #929, opened 2021-07-30
Add PHP 8.1 support

javer:


ramunasd:
Can You explain all the changes and why they are required for PHP8.1?

PhpAmqpLib/Wire/AMQPWriter.php
```diff
@@ -302,6 +308,12 @@ public function write_shortstr($s)
      */
     public function write_longstr($s)
     {
+        if ($s === null) {
+            $this->write_long(0);
+
+            return $this;
+        }
+
         $this->write_long(mb_strlen($s, 'ASCII'));
```
javer, line 317: Failed test: https://github.com/javer/php-amqplib/runs/3210161338?check_suite_focus=true#step:11:30
```
1) PhpAmqpLib\Tests\Unit\Message\AMQPMessageTest::serialize_properties with data set #9 (array(array('')), array(array(array('S', null))))
mb_strlen(): Passing null to parameter #1 ($string) of type string is deprecated

/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:305
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:435
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:365
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:383
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Message/AMQPMessage.php:529
/home/runner/work/php-amqplib/php-amqplib/tests/Unit/Message/AMQPMessageTest.php:21
```

Since PHP 8.1 passing null to non-nullable internal function parameters is deprecated: https://php.watch/versions/8.1/internal-func-non-nullable-null-deprecation

PhpAmqpLib/Wire/IO/StreamIO.php
```diff
@@ -344,6 +344,10 @@ protected function do_select($sec, $usec)
         $write = null;
         $except = null;
 
+        if ($sec === null && PHP_VERSION_ID >= 80100) {
+            $usec = null;
+        }
+
         return stream_select($read, $write, $except, $sec, $usec);
```
javer, line 351: Failed test: https://github.com/javer/php-amqplib/runs/3210161338?check_suite_focus=true#step:11:55
```
ErrorException: stream_select(): Argument #5 ($microseconds) should be null instead of 0 when argument #4 ($seconds) is null

/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php:347
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/IO/AbstractIO.php:88
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPReader.php:129
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPReader.php:155
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPReader.php:103
...
```

Since PHP 8.1 signature of the function `stream_select` has been changed: https://github.com/php/php-src/blob/master/ext/standard/basic_functions.stub.php#L1215
```php
function stream_select(?array &$read, ?array &$write, ?array &$except, ?int $seconds, ?int $microseconds = null): int|false {}
```

PhpAmqpLib/Wire/AMQPAbstractCollection.php
```diff
@@ -467,26 +471,31 @@ final public static function getDataTypeForSymbol($symbol)
         return $symbols[$symbol];
     }
 
+    #[\ReturnTypeWillChange]
     public function current()
```
javer, line 475: Failed test: https://github.com/javer/php-amqplib/runs/3210150874?check_suite_focus=true#step:11:13
```
PHP Fatal error:  During inheritance of Iterator: Uncaught Return type of PhpAmqpLib\Wire\AMQPAbstractCollection::current() should either be compatible with Iterator::current(): mixed, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPAbstractCollection.php:470
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPAbstractCollection.php:12
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:365
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:383
/home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Message/AMQPMessage.php:529
/home/runner/work/php-amqplib/php-amqplib/tests/Unit/Message/AMQPMessageTest.php:21
 in /home/runner/work/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPAbstractCollection.php on line 12
```

Since PHP 8.1 many internal classes have got correct return types: https://wiki.php.net/rfc/internal_method_return_types
And we cannot specify all of them explicitly, because this package supports PHP 7.0 which doesn't support all necessary return types. So as a workaround till PHP 9.0 we can use `ReturnTypeWillChange` attribute which suppresses this error.

The same reason for another the same changes.

.github/workflows/phpunit.yml
```diff
@@ -37,9 +37,13 @@ jobs:
             phpseclib: '^3.0'
             composer: 'composer:v2'
             coverage: none
+          - php: '8.1'
+            phpseclib: '3.0.x-dev'
```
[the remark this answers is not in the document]
ramunasd: Let's wait for compatible phpseclib/phpseclib version.

.github/workflows/phpunit.yml
```diff
@@ -70,7 +74,11 @@ jobs:
       run: php tests/phpinfo.php
 
     - name: Start broker service
-      run: docker start ${{ job.services.rabbitmq.id }}
+      run: docker restart ${{ job.services.rabbitmq.id }}
+
+    - name: Configure for PHP 8.1
+      run: composer config platform.php 8.0.99
+      if: matrix.php == '8.1'
```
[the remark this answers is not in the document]
ramunasd: Let's wait for phpspec/phpspec compatible version. There is no rush for PHP8.1 as it wasn't released yet.

PhpAmqpLib/Wire/IO/StreamIO.php
ramunasd: I would suggest to fix this for all versions. It makes no sense to use microseconds with non null seconds.

PhpAmqpLib/Wire/AMQPAbstractCollection.php
ramunasd: What is minimum required versions to make definitions compatible?

.github/workflows/phpunit.yml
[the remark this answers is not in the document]
javer: OK, let's wait, but not too long. Please note that real projects contains hundreds of dependencies, and all of them should be fixed and tested, as well as the projects themself. It's a good practice to be ready for the new PHP version **before** its release, not after. Otherwise the library becomes a blocker for everyone who uses it.

PhpAmqpLib/Wire/IO/StreamIO.php
javer: Unfortunately it won't work, because previous versions of PHP don't accept null for `microseconds` argument, see https://github.com/php/php-src/blob/php-8.0.9/ext/standard/basic_functions.stub.php#L1198
```php
function stream_select(?array &$read, ?array &$write, ?array &$except, ?int $seconds, int $microseconds = 0): int|false {}
```

PhpAmqpLib/Wire/AMQPAbstractCollection.php
javer: As we can see from the error message `should either be compatible with Iterator::current(): mixed` it requires `mixed` type which was introduced in PHP 8.0.

javer:
@ramunasd I've pushed changes with workaround removal, because all dependencies are now supports PHP 8.1. PHP 8.1 RC1 is also released. Maybe it's time to start testing against it?

PhpAmqpLib/Wire/AMQPWriter.php
ramunasd: I guess a unit test is missing :)

javer: @ramunasd I've added unit test for these cases and fixed the logic.

PR #942, opened 2021-09-23
Returned connection can be null

irinikp:
Method getConnection may return `null`, therefore PHPDocs should be changed accordingly.

### DIFF
--- a/PhpAmqpLib/Channel/AbstractChannel.php
+++ b/PhpAmqpLib/Channel/AbstractChannel.php
@@ -162,7 +162,7 @@ abstract class AbstractChannel
     }
 
     /**
-     * @return AbstractConnection
+     * @return AbstractConnection|null
      */
     public function getConnection()
     {


PhpAmqpLib/Channel/AbstractChannel.php
```diff
@@ -162,7 +162,7 @@ public function setBodySizeLimit($max_bytes)
     }
 
     /**
-     * @return AbstractConnection
+     * @return AbstractConnection|null
```
andrew-demb, line 165: Connection property cannot be null.
Phpdoc over property are incorrect and should fixed.

andrew-demb (changes requested):

PhpAmqpLib/Channel/AbstractChannel.php
ramunasd: Under some rare circumstances connection property can be null. Usually You wont get `null` from this method. So there is no need to fix anything.

andrew-demb: @ramunasd Didn't know that before. I think we can (and should) add throw exception (in `getConnection`, `next_frame`, `prepare_method_frame`) on null connection to be safe on typing level?

Exception like this: https://github.com/php-amqplib/php-amqplib/blob/3f426df3ee8adcec14db006c1265e59858437be3/PhpAmqpLib/Channel/AbstractChannel.php#L245

ramunasd: AFAIK all public methods already has check for existing connection. Perhaps some left without checks. I see no reason why we should add this check for internal(non-public) methods, but for `getConnection` `next_frame` we can add call to `checkConnection()`

irinikp: [Here](https://github.com/php-amqplib/php-amqplib/blob/3f426df3ee8adcec14db006c1265e59858437be3/PhpAmqpLib/Channel/AbstractChannel.php#L57) you declare the variable as `/** @var null|AbstractConnection */`

And [here](https://github.com/php-amqplib/php-amqplib/blob/3f426df3ee8adcec14db006c1265e59858437be3/PhpAmqpLib/Channel/AbstractChannel.php#L244) you check if the `$this->connection` is `null`. and throw the `AMQPChannelClosedException` exception you are suggesting, because it makes sense to throw it when someone needs to call `send_method_frame`. But does it make sense to throw an Exception when you simply need to get the connection?

irinikp: In a scenario where I use this package and I simply need to check if I have a connection, I need to call `getConnection`, not any other method extending this class. But then there's an inconsistency between the returned type and what is expected (a possibility of `null` is expected).

ramunasd: @irinikp I wonder, why do You need connection object? What does it gives to You?
Despite inconsistency in docblock, it does not prevent You from getting and checking connection.

irinikp: No it does not prevent me. But the inconsistency in the docblock is thrown as an error in our code-quality tests, and I have to cast the output as /** @var AbstractConnection|null $amqpConnection */

irinikp: And to answer the question of why do I need to check the connection object, I need to check if it is `null` (which means that the connection is closed) and if that's the case I follow some steps that are not related to other methods of your package. 

Therefore I end-up with something like this:
```
$amqpConnection = $libChannel->getConnection();
if ($amqpConnection === null ) {
 // steps
}
```
this `if` check doesn't make sense for any code-quality check tool when the output of `getConnection` is never `null` (which is not the case really)

ramunasd: I can suggest using other method - `$channel->is_open()` . It will return false if either connectionor channel was closed. Does this fit your need better? @irinikp

irinikp: If `$channel->is_open()` is the same thing as `$this->connection === null`, then why do you check both [here](https://github.com/php-amqplib/php-amqplib/blob/3f426df3ee8adcec14db006c1265e59858437be3/PhpAmqpLib/Channel/AMQPChannel.php#L177)?

andrew-demb: @ramunasd I can add one more argument for the fix: new contributors can initiate PR's with added return types everywhere based on information from phpdoc (or any automatic refactoring tool can do the same).
Misleading phpdoc will lead to trigger `TypeError` if a reviewer weren't known these internal details and prevent the code from breaking the staff.

ramunasd: @irinikp many people worked on this library, each trying to improve and fix things. Now, these direct questions like "why do you do something" are unacceptable.

If something was done, it means there was the reason. Nobody here are in charge of answering unrelated questions. Just use commit history to find answers.

irinikp: It was a rhetoric question to point out that if the checks `$channel->is_open()` and `$this->connection === null` were the same thing, you wouldn't have added them both on that line of code.

PR #957, opened 2021-11-10
Fix docblocks as per PHPStan

dereuromark:
as per

    $delivery_tag = $reader->read_longlong();
    
which returns `int|string`, can always be both it seems

our reports were
```                                                                           
  192    Parameter #1 $delivery_tag of method PhpAmqpLib\Channel\AMQPChannel::basic_ack() expects int, string given.       
  203    Parameter #1 $delivery_tag of method PhpAmqpLib\Channel\AMQPChannel::basic_reject() expects int, string given.    
```

### DIFF
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -776,7 +776,7 @@ class AMQPChannel extends AbstractChannel
     /**
      * Acknowledges one or more messages
      *
-     * @param int $delivery_tag
+     * @param int|string $delivery_tag
      * @param bool $multiple
      */
     public function basic_ack($delivery_tag, $multiple = false)
@@ -830,7 +830,7 @@ class AMQPChannel extends AbstractChannel
     /**
      * Handles the deletion of messages from this->publishedMessages and dispatches them to the $handler
      *
-     * @param int $delivery_tag
+     * @param int|string $delivery_tag
      * @param bool $multiple
      * @param callable $handler
      */
@@ -877,7 +877,7 @@ class AMQPChannel extends AbstractChannel
     /**
      * Rejects one or several received messages
      *
-     * @param int $delivery_tag
+     * @param int|string $delivery_tag
      * @param bool $multiple
      * @param bool $requeue
      */
@@ -890,7 +890,7 @@ class AMQPChannel extends AbstractChannel
     /**
      * Ends a queue consumer
      *
-     * @param string $consumer_tag
+     * @param int|string $consumer_tag
      * @param bool $nowait
      * @param bool $noreturn
      * @throws \PhpAmqpLib\Exception\AMQPTimeoutException if the specified operation timeout was exceeded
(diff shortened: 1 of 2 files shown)


ramunasd:
This library isn't strongly typed and cannot be. Please take delivery tag from AMQPMessage where it is always `int`. Then your static analysis will be fine.

dereuromark:
The null part is also wrong, please then fix up that part

ramunasd:
We will, definitely, thanks. That's a long iterative process :)

PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -513,7 +513,7 @@ public function prepare_content($channel, $class_id, $weight, $body_size, $packe
      * @param int $channel
      * @param array $method_sig
      * @param AMQPWriter|string $args
-     * @param null $pkt
+     * @param AMQPWriter|null $pkt
```
dereuromark, line 516: This part should be merged as PHPStan correctly calls this invalid as just null.

PhpAmqpLib/Channel/AMQPChannel.php
```diff
@@ -890,7 +890,7 @@ public function basic_nack($delivery_tag, $multiple = false, $requeue = false)
     /**
      * Ends a queue consumer
      *
-     * @param string $consumer_tag
+     * @param int|string $consumer_tag
```
dereuromark, line 893: Also these are weird. As you just said, you are using the "int" version.

PhpAmqpLib/Channel/AMQPChannel.php
```diff
@@ -951,7 +951,7 @@ public function is_consuming()
      * @link https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.consume
      *
      * @param string $queue
-     * @param string $consumer_tag
+     * @param int|string $consumer_tag
```
dereuromark, line 954: Same

ramunasd:
Thanks @dereuromark for Your review. Actually we are aware of all those inconsistencies. It's quite easy to check everything with any static analysis tool. And it will show even more warnings :)
The thing is that every change in doc blocks or type hints can be considered as breaking change. That's why we prefer to fix as much as possible and release those fixes only with next major version.

Thanks again. We will take care of it.

dereuromark:
No, those are not breaking changes but fixing documentation of obvious lies.

andrew-demb:
> The thing is that every change in doc blocks or type hints can be considered as breaking change. That's why we prefer to fix as much as possible and release those fixes only with next major version.

But if you make the input type wider (from `null` to `AMQPWriter|null`, or `string` to `int|string`) it is won't be considered as a BC break from a user perspective (see also PHP type covariance https://www.php.net/manual/en/language.oop5.variance.php).
Only descendants of changed classes can be 'affected' (they don't expect more types), but they are already 'broken' because 'invalid' types are already passed from several places.

PR https://github.com/php-amqplib/php-amqplib/pull/948 with similar changes (https://github.com/php-amqplib/php-amqplib/pull/948/files#diff-a79aae4447c58094ac8a6ac96a7ead509c19c5cc60d58c1746aae6092f9d32c4R172) was merged before, isn't this the same?

dereuromark:
I am talking about docblocks, not functional changes!

PR #971, opened 2022-01-27
Custom PCNTL Heartbeat Sender

khepin:
The existing `PCNTLHeartbeatSender` implementation uses `SIGALARM`
while convenient, for that use case, it also clashes with other libraries
that also want to use that same signal.

In our case, Laravel uses `SIGALARM` to define a timeout for job execution
on queue workers. Also a valid use case for it. But having both libraries
define their own signal handler for the same signal and overwrite each other's
alarm timeout makes a big mess where nothing works.

This allows defining a `pcntl_async_signals` handler that works with
a different signal than `SIGALARM`. The specific signal can be chosen at
construction time and is set to `SIGUSR1` by default.

The code uses `pcntl_fork` to create a separate process that sends the
chosen signal at regular intervals.

### DIFF
--- a/PhpAmqpLib/Connection/Heartbeat/AbstractSignalHeartbeatSender.php
+++ b/PhpAmqpLib/Connection/Heartbeat/AbstractSignalHeartbeatSender.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace PhpAmqpLib\Connection\Heartbeat;
+
+use PhpAmqpLib\Connection\AbstractConnection;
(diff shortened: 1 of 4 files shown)


ramunasd:
Thanks for alternative approach. This indeed helps in some cases.
Can You extract same parts from both heartbeat checker into abstract class or trait? I see we have code duplication here.

khepin:
Refactored to extract some parts into an abstract class.

ramunasd:
Nice, thank You

ramunasd (approved):

PR #975, opened 2022-02-09
unified configuration class and factory for all kinds of connection

ramunasd:
The aim of this PR is to put all connection configuration options into one object and reuse it for all kind of connections. Having this, we can improve several things:
- get rid of long argument list in each connection class
- unify default values
- deprecate global constants - once defined, they cannot be changed
- useable in applications with dependency injection(DI)

For now, previous approach takes precendence. Class arguments and global constants over connection config. This wil be canged in next major version.


lukebakken (approved):
This is great! Thank you!

PR #976, opened 2022-02-14
Propagate real exceptions

dmitryuk:
The pr adds real exceptions to target method `AbstractConnection::channel`.

Developer will can understand transparently that method `channel` can throw `AMQPTimeoutException,AMQPConnectionClosedException ` and do `reconnect`


PhpAmqpLib/Channel/AMQPChannel.php
```diff
@@ -271,8 +274,10 @@ protected function channel_flow_ok($reader)
 
     /**
      * @param string $out_of_band
-     * @throws \PhpAmqpLib\Exception\AMQPTimeoutException if the specified operation timeout was exceeded
-     * @return mixed
```
ramunasd, line 275: Actually it returns something, please leave it here.

PR #980, opened 2022-03-02
Channel cannot be null

dmitryuk:
Phpstan confused that `channel_id` cannot be null in any way, because it has been set in constructor

```php
public function __construct(AbstractConnection $connection, $channel_id)
    {
        $this->connection = $connection;
        $this->channel_id = (int)$channel_id;
...
}
```


ramunasd:
Thanks for suggestion. Actually this property can be null.
https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Channel/AMQPChannel.php#L144

PR #983, opened 2022-03-10
Throw connection exceptions on batch publish

foment:
Hello,

in bulk publishing, the connection existence is not checked (PhpAmqpLib/Channel/AMQPChannel.php:1232 in publish_batch method) and therefore we can't catch proper exception, like for example AMQPChannelClosedException.

This sometimes happens when there is a delay between opening the connection and actual publishing and for example in the meantime connection was closed for any reason whatsoever.

Therefore, here is the suggestion to throw a proper exception already in the prepare_method_frame method, like it's done for send_method_frame method in the AbstractChannel class.

### DIFF
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -1235,6 +1235,8 @@ class AMQPChannel extends AbstractChannel
             return;
         }
 
+        $this->checkConnection();
+
         /** @var AMQPWriter $pkt */
         $pkt = new AMQPWriter();
 
@@ -1265,7 +1267,6 @@ class AMQPChannel extends AbstractChannel
             }
         }
 
-        $this->checkConnection();
         $this->connection->write($pkt->getvalue());
         $this->batch_messages = array();
     }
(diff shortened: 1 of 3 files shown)


ramunasd:
Hi @foment , thanks for pointing this out and your efforts. Can You explain why You picked method `prepare_method_frame` ? This one is called only from `prePublish()` and only when "publish cache" is not warmed up.
Isn't better to check connection in `publish_batch()` ? Well, there is one check, maybe it's better to move it to beginning of method?
Anyways, this check does not garantuee stable connection. I would like to suggest enabling keepalive and heartbeat with signal senders. They are doing great job when scripts aren't responsive enough.

foment:
Hi @ramunasd,

thank you for the quick reply.

The idea behind throwing an exception in `prepare_method_frame` is more to be aligned with the current implementation like in `send_method_frame` for example, where connection is checked for it's existence. You're right, as an alternative, I could move `$this->checkConnection();` call in the `publish_batch` method to the beginning (before actual call to `prePublish` method).

One of the reasons behind all of this is to be able to catch proper exception in one of the legacy projects I'm currently working on and in general to have proper connection check before actually using it (like in `publish_batch`), which leads to php fatal error.

ramunasd:
So I don't see reason why "packet" method should be aware about connection. Please follow my suggestion. Thanks

foment:
PR updated with implemented suggestion.

ramunasd:
We still use older version of phpunit :)

foment:
Fix is on the way :)

PR #1013, opened 2022-07-27
Fix StreamIO ssl_protocol handling

frikiluser:
Before this changes AMQPSSLConnection class only connect using ssl when $ssl_options is non empty array.

This commit fixes that situation.

While this request is being reviewed, please avoid the widely recommended option:
```
['verify_peer' => false]
```

Instead use any of PHP SSL context options [0].
I suggest 'verify_peer' with the default value (true):
```
['verify_peer' => true]
```

Best regards,

[0] https://www.php.net/manual/en/context.ssl.php


frikiluser:
I think it's broken since cce2d96

ramunasd:
Hi, thanks for Your efforts. Can You describe the problem you are trying to solve?
Also, tests for your use case are missing.

frikiluser:
Hi @ramunasd ,

I would expect all connections using AMQPSSLConnection are SSL.

With the current `php-amqplib` code it doesn't work that way. By default it uses plain TCP connections even when using AMQPSSLConnection class. To enable real SSL connection the user should set a $ssl_options (PHP SSL context options).

This line opens a plain TCP connection (👍🏾):
```
$connection = new AMQPStreamConnection('myhost', 1234, 'user', 'pass');
```

This line opens a plain TCP connection (😿):
```
$connection = new AMQPSSLConnection('myhost', 1234, 'user', 'pass');
```

This line opens a SSL connection (👍🏾):
```
$connection = new AMQPSSLConnection('myhost', 1234, 'user', 'pass','/',['verify_peer'=>true]);
```


After merging this pull request, this line will open a SSL connection (👍🏾🥳):
```
$connection = new AMQPSSLConnection('myhost', 1234, 'user', 'pass');
```


Thanks for your time!

ramunasd:
In the future there will be no different connection classes. The recommended way is to use connection factory which is responsible for creating right class for you. However we still support old approach.

Please create test case for this fix.

lukebakken (changes requested):
Test(s) are required, thank you.

PR #1034, opened 2022-10-14
Change the version that is displayed in the properties of the client

i3bepb:
displayed in the properties of the client


ramunasd:
It will be different on next release. Thanks anyways

PR #1063, opened 2023-01-13
Optimize frame parsing logic

ramunasd:
This PR improves the speed and simplicity of binary message parser. Historically we use mb string functions for splitting and extracting parts of binary stream. But they can be replaced by unpack(), moreover, several calls can be combined into one.

In overall this PR has:
2 less in-memory read buffers
5 less mb_* calls
4 less unpack() calls
faster SplQueue instead of array for frame queue
less temporary arrays
constants instead of hardcoded strings for readability
no functional changes

When measuring with `make benchmark` it shows
~15% better performance using my macbook pro 2,3 GHz Quad-Core Intel Core i7, PHP 7.4
~17% using 32bit raspberry pi 3, PHP 7.3

### DIFF
--- a/PhpAmqpLib/Channel/AbstractChannel.php
+++ b/PhpAmqpLib/Channel/AbstractChannel.php
@@ -35,9 +35,9 @@ abstract class AbstractChannel
 
     /**
      * Lower level queue for frames
-     * @var array
+     * @var \SplQueue|Frame[]
      */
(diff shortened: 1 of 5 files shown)


lukebakken (approved):
This is _VERY_ impressive! Thanks!!!

michaelklishin:
@ramunasd thank you for your non-trivial contributions!

Issue #1080, opened 2023-03-28
Positive heartbeat of client is overwritten by server's 0 heartbeat value. All versions from 3.3.1 affected


trb-dev:
**Issue:** Heartbeat doesn't work properly as described in rabbitmq documentation. Main problem that positive client's heartbeat is overwritten by server's 0 value.

**Version(s) affected:** all from v3.3.1 (v.3.3.0 version had a good try, but you implemented heartbeat disabling by client side fix on v.3.3.1)

**Code place:** https://github.com/php-amqplib/php-amqplib/blob/703eef998946f4972e630c3e3d827f7cf020fe68/PhpAmqpLib/Connection/AbstractConnection.php#L947-L950

**Previous similar issue fixes:**  
- https://github.com/php-amqplib/php-amqplib/issues/1018 - this fix was mentioned as broken v3.3.0 due to not being able to disable heartbeat by client
- https://github.com/php-amqplib/php-amqplib/pull/1030 - v3.3.1 hotfix for 3.3.0 and current negotation version until now

---------------------

Rabbitmq heartbeat negotation rules
---------------------

> If either value is 0 (see below), the greater value of the two is used
Otherwise the smaller value of the two is used
A zero value indicates that a peer suggests disabling heartbeats entirely. To disable heartbeats, both peers have to opt in and use the value of 0. This is highly recommended against unless the environment is known to use [TCP keepalives](https://www.rabbitmq.com/heartbeats.html#tcp-keepalives) on every host.

Example 1. Fail - rule **If either value is 0 (see below), the greater value of the two is used** :
---------------------
- Servers heartbeat config is default - 60
- Client heartbeat value - 0
  - Heartbeat stays as 0 due to code if statement `if ($this->heartbeat > 0) {`

Example 2. Fail - rule **If either value is 0 (see below), the greater value of the two is used** :
---------------------
- Servers heartbeat config is 0
- Client heartbeat value - 75
  - Server's heartbeat overwrites client's heartbeat and it stays as 0 due to code `min($this->heartbeat, $v);`

Example 3. Success - rule **Otherwise the smaller value of the two is used** :
---------------------
- Servers heartbeat config is default - 60
- Client heartbeat value - 75
  - Servers heartbeat overwrites clients heartbeat as it has lower value and connection is left with value of 60. 

---------------------

One of the code suggestions could be to meet their described negotations algorithm:

```
        $serverHeartBeat = (int) $args->read_short();
[...]

PR #1081, opened 2023-03-30
Fix server heartbeat overwrite client value

ramunasd:
This is fix for a case described in #1080 as Example no. 2


lukebakken (approved):
Thanks! FYI I'm away from work until April 10th.

Issue #1094, opened 2023-06-05
`AMQPTimeoutException` occurs during socket interruptions


mrtus:
**Version(s) affected**: v3.5.2

**Description**  
When a SIGNAL occurs, the exception `AMQPTimeoutException` is thrown at `AMQPIOReader::wait::102`
Since the channel technically did not timeout, but was interrupted... It is hard to differenciate between stream interruptions and those timeouts, which could be a symptom of something completely different.

We rely on scaling our workers based on CPU usage & available messages on the queue. Whenever consumers are being scaled down a SIGNAL is sent, we try to handle those but of course when a connection is open and interupted... the situation becomes a little bit harder.

**How to reproduce**  
Have a channel waiting for a message
Send a SIGNAL to interrupt the stream
Exception is thrown

**Possible Solution**  
Instead of throwing a "timeout" exception, I would like to propose to instead throw a "AMQPConnectionInterruptedException` or something a-like.

We can most likely throwing such an exception based on calculating if the timeout is really a timeout (the time a wait took is longer than the given timeout param). Or am I missing some insights here?

Let me hear your thoughts on this issue.
Or if there would be a better solution to handle a SIGNAL?

Thank you

ramunasd:
I guess there might be another exception for this case, something like `AMQPWaitInterruptedException`. But we can introduce it only in next major version, because it can be considered as breaking change.

For now You can use signal handler to save some kind of flag about interruption. When you have it, then AMQPTimeoutException can be handled differently.

mrtus:
Hi @ramunasd thank you for your reply

> I guess there might be another exception for this case, something like `AMQPWaitInterruptedException`. But we can introduce it only in next major version, because it can be considered as breaking change.

That would be awesome, yes!
An alternative solution could be to extend `AMQPWaitInterruptedException` with the existing `AMQPTimeoutException`, but maybe that is cheating versioning and entirely not correct 😛.

> For now You can use signal handler to save some kind of flag about interruption. When you have it, then AMQPTimeoutException can be handled differently.

That is exactly how I resolved the use-case for now, and it is working nicely too.

matheusjohannaraujo:
Is this the error you are having???

I'm having this error when I go to read the queue, basically every time the anonymous sub callback takes more than 10 seconds to execute, this disconnect happens on localhost.

```php
[...]

PR #1121, opened 2023-10-05
Make sure, that a select interruped by a signal is retried

mruell:
Fixes https://github.com/php-amqplib/php-amqplib/issues/1094 and https://github.com/php-amqplib/php-amqplib/issues/1118

### DIFF
--- a/PhpAmqpLib/Exception/AMQPIOWaitInterruptedException.php
+++ b/PhpAmqpLib/Exception/AMQPIOWaitInterruptedException.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace PhpAmqpLib\Exception;
+
+class AMQPIOWaitInterruptedException extends AMQPRuntimeException
+{
+}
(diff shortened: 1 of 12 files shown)


mruell:
Alright, I think the PR is now in a merge ready state
Happy to get some feedback :)

ramunasd:
Thanks for Your attempt to fix it. Very likelly we will use another approach. The main problem is in misleading AbstractIO::select() result and wrong exception, not the fact it was interrupted.

intxcc:
Hmm - so it is supposed to be interrupted?

mruell:
Thanks for the constructive feedback!
I changed the PR to instead throw the new Exception `AMQPIOWaitInterruptedException`

Hope that helps :)

Best,
Marvin

mruell:
Just noticed what you mean - the error_handler does already catch the interrupt, but does ignore it
So the signal catcher is not needed

Thus I'll close this PR
At least learned a lot about the amqplib ^^

If you want I can create a PR for throwing the correct error within the error_handler

PR #1134, opened 2023-10-23
Fix deprecation warning

frederikbosch:
func_num_args() === 17 is always true because the optional arguments


ramunasd:
@egorgrushko WDYT?

egorgrushko:
@ramunasd This will work. However, if someone will call the constructor, passing 17 parameters explicitly (with $ssl_protocol = null), there will be no deprecated error, although it should be since passing 17 parameters is actually deprecated.
And func_num_args does NOT count optional parameters if they are not passed explicitly.

So I don't think these changes are correct.

PR #1145, opened 2023-12-01
ipv6 support

InnoT20:
**Description of Changes**
This pull request adds support for IPv6 for socket connection

**Changes Made**
- Added IPv6 support for working with sockets.
- Implemented domain DNS checking to handle cases where the domain leads to an IPv6 address.

### DIFF
--- a/PhpAmqpLib/Wire/IO/SocketIO.php
+++ b/PhpAmqpLib/Wire/IO/SocketIO.php
@@ -15,8 +15,6 @@ class SocketIO extends AbstractIO
     /** @var null|resource */
     private $sock;
 
-    private $isIpv6 = false;
-
     /**
      * @param string $host
      * @param int $port
@@ -44,7 +42,6 @@ class SocketIO extends AbstractIO
         $this->initial_heartbeat = $heartbeat;
         $this->keepalive = $keepalive;
         $this->canDispatchPcntlSignal = $this->isPcntlSignalEnabled();
-        $this->isIpv6 = $this->isIpv6();
 
         /*
             TODO FUTURE enable this check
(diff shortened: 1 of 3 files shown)


PhpAmqpLib/Wire/IO/SocketIO.php
```diff
@@ -34,14 +36,15 @@ public function __construct(
...
-        $this->host = $host;
+        $this->host = str_replace(['[', ']'], '', $host);
         $this->port = $port;
         $this->read_timeout = (float)$read_timeout;
         $this->write_timeout = (float)($write_timeout ?: $read_timeout);
         $this->heartbeat = $heartbeat;
         $this->initial_heartbeat = $heartbeat;
         $this->keepalive = $keepalive;
         $this->canDispatchPcntlSignal = $this->isPcntlSignalEnabled();
+        $this->isIpv6 = $this->isIpv6();
```
ramunasd, line 47: We don't need to keep this variable as class property if it is being used in only one place and only once.

PhpAmqpLib/Wire/IO/SocketIO.php
```diff
@@ -60,7 +63,7 @@ public function __construct(
      */
     public function connect()
     {
-        $this->sock = socket_create(AF_INET, SOCK_STREAM, SOL_TCP);
+        $this->sock = socket_create(!$this->isIpv6 ? AF_INET : AF_INET6, SOCK_STREAM, SOL_TCP);
```
ramunasd, line 66: Call method `isIpv6()` instead of var.

tests/Unit/Wire/IO/StreamIOTest.php
```diff
@@ -49,4 +49,17 @@ public function select_must_throw_io_exception()
...
+     * @test
+     */
+    public function connect_ipv6()
+    {
+        $streamIO = new StreamIO(HOST6, PORT, 0.1, 0.1, null, false, 0);
+        $streamIO->connect();
+        $ready = $streamIO->select(0, 0);
+        $this->assertEquals(0, $ready);
+
+        return $streamIO;
```
ramunasd, line 63: There is no need to return anything from test method. Just make sure all assertions are implemented.

ramunasd (changes requested):

ramunasd:
Thanks for improvement!
I have question regarding stream implementation. Does it support ipv6 addresses without any modifications?

InnoT20:
> Thanks for improvement! I have question regarding stream implementation. Does it support ipv6 addresses without any modifications?

The stream implementation supports IPv6 addresses without requiring any additional modifications. 
I've locally verified this on PHP version 8.1, confirming that the stream works seamlessly with IPv6 addresses.

ramunasd (approved):

PR #1150, opened 2024-01-19
Lazy connection leaks StreamIO context resource even when closed

janedbal:
Reproduction test for #1152


ramunasd:
We won't merge failing tests :D Please work on full PR, including fix and all the tests.

Issue #1159, opened 2024-02-09
Better support for OAuth2 authorization backend


kratkyzobak:
RabbitMQ supports OAuth2 authorization, which has some challenges for clients to implement.

** OAuth2 in RabbitMQ **

For login, nothing is changed - OAuth token is passed instead of password with any username (username is read from token), so there is no need to upgrade library at all.

But for long running processes, there is problem. OAuth access tokens has it's expiration. This is usually something about few minutes to one hour, depending on concrete IdP. RabbitMQ developers [decided to "unauthorize"](https://www.rabbitmq.com/oauth2.html#token-expiration) connections after token expiration. **So long-running workers which are connected to RabbitMQ needs to upgrade access token in connection before expiration.**

** How it can be implemented **

First, generally simple step, is to expose any method to support `connection.update-secret` call for long-running processes.

This would be still very low level support for OAuth2 authorization as library consumer still needs to handle some timer function to upgrade secret.

I believe, there is few possibilities to handle this inside library, but I'm not sure what is "correct" (and would be accepted by maintainers). 

Simple part is initial password use. Instead of password, there can be some interface `TokenProvider` which would have method to return token and it's expiration - this is very common for OAuth2. Configuration and/or connection constructors would accept `null|string|TokenProvider $password` instead of just nullable strings. For initial connection it's easy - If `$password instanceof TokenProvider`, just call `provideToken` and use it as password. For lazy connection / reconnect - if token provider is present in instance, call provideToken before new connection is done. This part is still improvement from current state, as for long-running processes with lazy connections, where "password" can expire even before first connection attempt. (PR soon)

Hard part is when to call update-secret on existing connections. Library can either use some timer function OR enable user to use it's own timer and call library. In this library, there is already PCNTL/SIG handlers of heartbeats so it can be combined into something "better than just heartbeat", but it can also check access token expiration time and update it if possible. Problem is, that `pcntl_alarm` callback can be only one. So some refactoring would be needed to support both - heartbeats and updating secrets. Also, it would be good to try to support multiple RabbitMQ connections with heartbeats.

Another solution can be on library side, where check if "secret is valid" can be done during generic read/writes (or during heartbeats?).

Either variant, there should be code like:

```
if ($this->is_connected && $this->current_token->expiration < time() - $fewSeconds)) {
    $this->updatePassword($this->token_provider->provideToken());
}
```

[...]

PR #1160, opened 2024-02-09
Implement connection.update-secret to support rabbit_auth_backend_oauth2

kratkyzobak:
Implementation of "simple" part of https://github.com/php-amqplib/php-amqplib/issues/1159

This PR implements method `connection.update-secret` and exposes it to library consumers as `*Connection::updatePassword($password);`. Naming were chosen this way as it is just updating previously set `$password` in constructor. Method also updates internal password properties of Connection classes. This enables to reconnect and/or support lazy connections.


composer.json
```diff
@@ -36,6 +36,7 @@
     },
     "require-dev": {
         "ext-curl": "*",
+        "ext-openssl": "*",
```
ramunasd, line 39: Library itself does not need this extension, only tests.

PhpAmqpLib/Helper/Protocol/Protocol091.php
```diff
@@ -171,6 +171,29 @@ public static function connectionUnblocked(AMQPReader $reader)
...
+        return array(10, 70, $writer);
+    }
+
+    /**
+     * @param AMQPReader $reader
+     * @return array
+     */
+    public static function connectionUpdateSecretOk(AMQPReader $reader)
+    {
+        $response = array();
```
ramunasd, line 193: Why not just `return [];` ?

PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -316,6 +316,38 @@ public function reconnect()
...
+            $this->login_response = implode("\0", $login_response_parts);
+        }
+
+        if ($this->is_connected) {
+            // send new secret to broker if currently connected
+            $this->x_update_secret($password);
+        }
+
+        // for potential cloning
+        $this->replace_password_in_construct_params($password);
```
ramunasd, line 340: Constructor params are deprecated and will be dropped in version 4.0

PhpAmqpLib/Connection/AMQPStreamConnection.php
```diff
@@ -131,4 +131,12 @@ protected static function try_create_connection($host, $port, $user, $password,
             $channel_rpc_timeout
         );
     }
+
+    /**
+     * @param string $password
+     */
+    protected function replace_password_in_construct_params($password)
```
ramunasd, line 138: pointless method

tests/generate_jwt_keys.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+// if not set, generates just two different keys, both for about 2 hours
+define('JWT_TOKEN_1', getenv('TEST_RABBITMQ_JWT_TOKEN_1') ? getenv('TEST_RABBITMQ_JWT_TOKEN_1') : generateJwtToken(7200));
+define('JWT_TOKEN_2', getenv('TEST_RABBITMQ_JWT_TOKEN_2') ? getenv('TEST_RABBITMQ_JWT_TOKEN_2') : generateJwtToken(7205));
+
+function generateJwtToken($expire = 7200)
```
ramunasd, line 7: this should be part of test, abstract test class or something "on demand" but not defined as global constats, whats the point?

PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -316,6 +316,38 @@ public function reconnect()
         $this->connect();
     }
 
+    /**
+     * Sets new password for this connection.
+     * Should be used for RabbitMQ's OAuth2 authentication backend to update current token.
+     * @param string $password
+     */
+    public function updatePassword($password): void
```
ramunasd, line 324: missing type hint

ramunasd (changes requested):

ramunasd:
Next minor version is only fo bugfixes and deprecations. Afterwards we will refactor many things related to connections. So we can't accept this for 3.x series. At best for 4.0.

https://github.com/php-amqplib/php-amqplib/discussions/1123

PhpAmqpLib/Helper/Protocol/Protocol091.php
kratkyzobak: Method is generated by your code.

composer.json
kratkyzobak: Hence require-dev.

PhpAmqpLib/Connection/AbstractConnection.php
kratkyzobak: When will version 4.0 released? Will version 4.0 handle OAuth?

PR #1163, opened 2024-02-15
Trigger ssl_protocol deprecation when it is really not null and not AMQPConnectionConfig instance

SergeyKhrystenko:
As for PHP 8.0+ when using named params, func_num_args always return number of all params, because php de-facto inserts default values for that variables. This always generates deprecation warning.
In this PR I've deleted check for arguments number and leave only check for not null or AMQPConnectionConfig instance. That's enough for triggering deprecation.
For not named arguments nothing changes

### DIFF
--- a/PhpAmqpLib/Connection/AMQPStreamConnection.php
+++ b/PhpAmqpLib/Connection/AMQPStreamConnection.php
@@ -45,7 +45,7 @@ class AMQPStreamConnection extends AbstractConnection
         $ssl_protocol = null,
         ?AMQPConnectionConfig $config = null
     ) {
-        if (func_num_args() === 17 || ($ssl_protocol !== null && $ssl_protocol instanceof AMQPConnectionConfig === false)) {
+        if ($ssl_protocol !== null && $ssl_protocol instanceof AMQPConnectionConfig === false) {
             trigger_error(
                 '$ssl_protocol parameter is deprecated, use stream_context_set_option($context, \'ssl\', \'crypto_method\', $ssl_protocol) instead (see https://www.php.net/manual/en/function.stream-socket-enable-crypto.php for possible values)',
                 E_USER_DEPRECATED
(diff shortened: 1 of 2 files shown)


PhpAmqpLib/Connection/AMQPStreamConnection.php
```diff
@@ -45,7 +45,7 @@ public function __construct(
         $ssl_protocol = null,
         ?AMQPConnectionConfig $config = null
     ) {
-        if (func_num_args() === 17 || ($ssl_protocol !== null && $ssl_protocol instanceof AMQPConnectionConfig === false)) {
+        if ($ssl_protocol !== null && $ssl_protocol instanceof AMQPConnectionConfig === false) {
```
ramunasd, line 48: this might be confusing, conditionA && conditionB === false, what is precedence?

SergeyKhrystenko: as mentioned here https://www.php.net/manual/en/language.operators.precedence.php first "instanceof" called and then comparison.
so this is the same as $ssl_protocol !== null && ($ssl_protocol instanceof AMQPConnectionConfig) === false

so we trigger error only when $ssl_protocol is string or smth else, but not an AMQPConnectionConfig instance

ramunasd: so You can use `!instanceof AMQPConnectionConfig` ?

SergeyKhrystenko: not exactly. we can use `!$ssl_protocol instanceof AMQPConnectionConfig` which is the same as `$ssl_protocol instanceof AMQPConnectionConfig === false`

SergeyKhrystenko:
is there any plan to introduce this change? using php version >= 8 generates a lot of deprecated logs

Issue #1181, opened 2024-06-11
first error message thrown by stream_socket_enable_crypto() is not propagated to caller


jjdunn:
**Version(s) affected**: 3.6.2
**PHP version**: 8.3.7
**RabbitMQ version**: 0.9.1

**Description**  
```stream_socket_enable_crypto()``` throws TWO errors if it cannot find the SSL certificate due to open_basedir restriction; but only the second of the two errors is propagated to the caller, making debugging more difficult than needed.

the first error is "stream_socket_enable_crypto(): open_basedir restriction in effect..." which is helpful. This is the real cause of the problem; but this error is NOT reported to the caller. it gets overwritten by the second error "unable to get real path of certificate file...." which is true; but not easy to debug

**How to reproduce**  
1. in php.ini, set open_basedir to a list of directories which includes the application, but does not include the path to SSL certificate and associated information (intermediate cert, keys, etc)
2. attempt to open an AMQPS connection
3. get the error "stream_socket_enable_crypto(): Unable to get real path of certificate file...."

**Possible Solution**  
\Wire\IO\AbstractIO::error_handler() should accumulate an array of errors (instead of just last_error); and ::throwOnError() would report the list of errors. I guess AbstractIO::last_error should be renamed as error_list or errors or similar. If you agree with this solution-design I would be happy to submit a PR...let me know.

**Additional context**  
this problem was not exposed in PHP 8.1.x and preceding versions

the open_basedir check when referencing local_cert and local_pk in was introduced in PHP 8.2:
https://www.php.net/ChangeLog-8.php
version 8.2.0:
Fixed bug https://github.com/php/php-src/issues/9310 (SSL local_cert and local_pk do not respect open_basedir).

see four attached screenshots:
1. stack trace at \Wire\IO\StreamIO::enableCrypto()
![stack-trace-1](https://github.com/php-amqplib/php-amqplib/assets/333222/5ecb88b5-53e8-4b61-a816-ab5a0ca86f14)

2. stack trace at \Wire\IO\StreamIO::error_handler()
![stack-trace-2](https://github.com/php-amqplib/php-amqplib/assets/333222/0429c29b-0303-442b-a454-55c97c1812d7)

3. StreamIO properties in first call to error_handler(), showing the error message which gets lost
![error-message-1-lost](https://github.com/php-amqplib/php-amqplib/assets/333222/a222f30f-4cf3-4648-8d29-b23b591ca8fd)

[...]

PR #1183, opened 2024-06-13
fix: create a new exception for each warning in IO layer with reference to previous

ramunasd:
As described in #1181 , only last warning message is exposed making debugging more difficult.
After this change each warning will be stored as exception with reference to previous. Last one won't have $previous.

All errors(in reverse order) can be extracted by iterating recursively over previous exceptions:
```php
$previous = $exception;
do {
$messages[] = $previous->getMessage();
$previous = $exception->getPrevious();
} wihle ($previous !== null);
```

### DIFF
--- a/PhpAmqpLib/Wire/IO/AbstractIO.php
+++ b/PhpAmqpLib/Wire/IO/AbstractIO.php
@@ -46,7 +46,7 @@ abstract class AbstractIO
     /** @var int|float */
     protected $last_write;
 
-    /** @var array<string, mixed>|null */
+    /** @var \ErrorException|null */
     protected $last_error;
 
     /** @var bool */
@@ -229,14 +229,10 @@ abstract class AbstractIO
 
     protected function throwOnError(): void
     {
-        if ($this->last_error !== null) {
-            throw new \ErrorException(
-                $this->last_error['errstr'],
-                0,
-                $this->last_error['errno'],
-                $this->last_error['errfile'],
-                $this->last_error['errline']
-            );
+        if ($this->last_error instanceof \ErrorException) {
+            $error = $this->last_error;
+            $this->last_error = null;
+            throw $error;
         }
     }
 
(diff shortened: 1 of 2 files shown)


jjdunn:
Hi Ramunas,

Thanks for taking on this issue!

I pulled the change and tried it on my system, after reverting open_basedir so it does not contain the SSL path

It works, but the suggested while loop to retrieve the nested errors is showing the last error twice:

```
        try {
            $MQClient = $this->openMQConnection();
        } catch (exception $e) {
            $error = '';
            do {
                $error .=  "\r\n"  . $e->getMessage();
                $e = $e->getPrevious();
            } while ($e !== null);
            DRLog::err($error);
            exit ("SpecialListIndex: could not connect via MQ.  Error: {$error}");
        }
```
outputs:
```
SpecialListIndex: could not connect via MQ.  Error: 
stream_socket_enable_crypto(): Unable to get real path of certificate file `<SNIP>.crt'
stream_socket_enable_crypto(): Unable to get real path of certificate file `<SNIP>.crt'
stream_socket_enable_crypto(): open_basedir restriction in effect. File(<SNIP>.crt) is not within the allowed path(s): (C:\Web;C:\windows\temp;)
```

this is certainly an improvement; but I'm not sure why the last error is there twice.

here's how it looks in the variable-inspector:
error-on-stack-twice

lukebakken (approved):

ramunasd:
@jjdunn maybe PHP tried more than path in setting `include_path` ?

jjdunn:
> @jjdunn maybe PHP tried more than path in setting `include_path` ?

no, the problem is that ```StreamIO::enableCrypto()``` is re-throwing the Exception, which adds the duplicate entry

\Wire\IO\StreamIO line 454 includes the current exception in the rethrown exception, which doubles it:
```
throw new AMQPIOException($exception->getMessage(), $exception->getCode(), $exception);
```

I'm not too bothered by this.  The new code is certainly an improvement in the problem reported in #1181.   Just pointing it out in case you want to further tidy. :-) 

1. screenshot in ```StreamIO::enableCrypto()```, after ```stream_socket_enable_crypto()``` failed due to ```open_basedir``` restriction.  
Note that there are _two_ Exceptions in the stack before the exception is re-thrown:
StreamIO-enableCrypto-catchException

2. screenshot in ```AbstractConnection::connect()``` in the catch block.  
Note there are now _three_ Exceptions in the stack.  
The topmost one got duplicated when the error was rethrown in ```enableCrypto()```:
AbstractConnection-connect-catchException

ramunasd:
Good catch, easy to fix :)

jjdunn:
> Good catch, easy to fix :)

Thanks.  I applied the patch (https://github.com/php-amqplib/php-amqplib/commit/f534e001ff425cc76b9c5a708d56e1b684ac9c9d) and verified on my system that it no longer logs the final exception twice.

As far as I'm concerned this is done.  Thanks !

PR #1195, opened 2024-09-28
fix: Mark channels closed when connection is closed

Xfaider48:
### This PR fixes this situiation:
When a connection gets broken (Got AMQPConnectionClosedException) we need to mark all opened channels as closed OR it can lead to some bugs.

### Tested versions
- 3.7.0
- 3.7.1

### Warning
I added a new call to the `\PhpAmqpLib\Connection\AbstractConnection::do_close`, but I'm sure that it's the most proper way to fix this problem.

### Steps to reproduce:
```php
$connection = new \PhpAmqpLib\Connection\AMQPStreamConnection(
    $host,
    $port,
    $user,
    $pass,
);

$firstChannel = $connection->channel();
$secondChannel = $connection->channel();

$firstChannel->queue_declare('test');
$firstChannel->basic_consume('test', '', false, true, false, false, function (\PhpAmqpLib\Message\AMQPMessage $message) {
    $message->ack();
});
try {
    $firstChannel->wait();
    // Emulate closing: force close connection using management plugin
} catch (\PhpAmqpLib\Exception\AMQPConnectionClosedException $exception) {
    // We got AMQPChannelClosedException here
    var_dump($exception->getMessage() === 'CONNECTION_FORCED - Closed via management plugin(0, 0)'); 
    // Output: bool(true)
}

$connection->reconnect();
// Try to use old channel (expecting an exception AMQPChannelClosedException 'Channel connection is closed.')
$secondChannel->basic_publish(new \PhpAmqpLib\Message\AMQPMessage('test'), '', 'test');

try {
    // Try to get 
    $channelAfterReconnect = $connection->channel();
} catch (\PhpAmqpLib\Exception\AMQPConnectionClosedException $exception) {
    var_dump($exception->getMessage());
    // Output: string(47) "CHANNEL_ERROR - expected 'channel.open'(60, 40)"
}
```

### DIFF
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -207,9 +207,19 @@ class AMQPChannel extends AbstractChannel
         ), false, $this->channel_rpc_timeout);
     }
 
-    public function markClosed()
+    /**
+     * Closes a channel if no connection or a connection is closed
+     *
+     * @return bool
+     */
+    public function closeIfDisconnected(): bool
     {
+        if (!$this->connection || $this->connection->isConnected()) {
+            return false;
+        }
+
         $this->do_close();
+        return true;
     }
 
     /**
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -207,6 +207,11 @@ class AMQPChannel extends AbstractChannel
         ), false, $this->channel_rpc_timeout);
     }
 
+    public function markClosed()
+    {
+        $this->do_close();
+    }
+
     /**
      * @param AMQPReader $reader
      * @throws AMQPProtocolChannelException
--- a/PhpAmqpLib/Connection/AbstractConnection.php
+++ b/PhpAmqpLib/Connection/AbstractConnection.php
@@ -424,6 +424,7 @@ abstract class AbstractConnection extends AbstractChannel
         $this->frame_queue = new \SplQueue();
         $this->method_queue = [];
         $this->setIsConnected(false);
+        $this->markChannelsClosed();
         $this->close_input();
         $this->close_socket();
     }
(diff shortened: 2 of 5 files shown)


PhpAmqpLib/Channel/AMQPChannel.php
```diff
@@ -207,6 +207,11 @@ public function close($reply_code = 0, $reply_text = '', $method_sig = array(0,
         ), false, $this->channel_rpc_timeout);
     }
 
+    public function markClosed()
```
ramunasd, line 210: There is no point of adding public wrapper for protected function. Also we don't want to have potentially harmful public method.
We should either make public `do_close()` or do something smarter. I suggest this way:
```
public function closeIfDisconnected(): void
{
    if ($this->connection && !$this->connection->isConnected()) {
        $this->do_close();
    }
}
```

PhpAmqpLib/Connection/AbstractConnection.php
```diff
@@ -424,6 +424,7 @@ protected function do_close()
         $this->frame_queue = new \SplQueue();
         $this->method_queue = [];
         $this->setIsConnected(false);
+        $this->markChannelsClosed();
```
ramunasd, line 427: word `mark` is confusing and seems we are faking something. Why not just `closeChannels()` ?

ramunasd (changes requested):

PhpAmqpLib/Channel/AMQPChannel.php
Xfaider48: Agree, I'll change this

PhpAmqpLib/Connection/AbstractConnection.php
Xfaider48: We already have `closeChannels()`, but we can't use it because we need to close "silently", without sending anything.

ramunasd (approved):

PR #1199, opened 2024-11-17
fix: Deprecate implicitly nullable types

mbutkereit:
This pull request explicitly flags all nullable parameter. 

https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


ramunasd:
fixed in #1200

PR #1208, opened 2025-02-18
Provide the amqp extension so composer is satisfied it is present

SiebeVE:
https://getcomposer.org/doc/04-schema.md#provide


ramunasd:
Seems this is related to https://github.com/php-amqp/php-amqp which is completely different piece of software

PR #1213, opened 2025-04-24
refactor: using [] instead of the array()

kiani01lab:
As of PHP 5.4, it is possible to use a short array syntax [] instead of array(). I think it's prettier and up-to-date


ramunasd:
Thanks for your PR. Only this changed file has 72 more old style arrays. I suggest to replace them all at once or none.

PR #1234, opened 2026-01-08
Replaced Httpful with Guzzle

fazed:
Hi,

A small pull request which replaces the Httpful library with Guzzle due some security concerns. Should not affect the workings of the module since it's only used in functional tests.


ramunasd:
Thanks for your efforts to make this lib better.
I was planning to drop support for php <8.0 and upgrade this library, effectively solving existing CVE's

PR #1238, opened 2026-03-17
Fix PHP 8.5 deprecation in basic_consume when consumer_tag is null

janedbal:
The `wait()` method can return `null` (e.g. when the while loop in `AbstractChannel::wait()` exits without dispatching). In `basic_consume`, this null value is then used as an array offset:

```php
$this->callbacks[$consumer_tag] = $callback;
```

This triggers the PHP 8.5 deprecation:

```
Deprecated: Using null as an array offset is deprecated, use an empty string instead
```

This adds a null check before using `$consumer_tag` as an array key.

### DIFF
--- a/PhpAmqpLib/Channel/AMQPChannel.php
+++ b/PhpAmqpLib/Channel/AMQPChannel.php
@@ -1024,7 +1024,9 @@ class AMQPChannel extends AbstractChannel
             ), false, $this->channel_rpc_timeout);
         }
 
-        $this->callbacks[$consumer_tag] = $callback;
+        if ($consumer_tag !== null) {
+            $this->callbacks[$consumer_tag] = $callback;
+        }
 
         return $consumer_tag;
     }


ramunasd:
Thanks, is this something that really happened in your setup or just static analysis catch?
I see just 3 lines above there is condition which ensures array is not empty. So how it can happen?

janedbal:
I'm sorry, my Claude Code got sidetracked while working on this issue. Force pushed proper fix of the real issue. Here is a callstack of such failure in our sentry:

image

ramunasd:
Then it's also interesting.
If `$nowait = true`, then `$consumer_tag` as argument must be non empty - this is checked on line 1000.
If `$nowait = false`, then `$consumer_tag` is retrieved from broker(auto generated). So in your case broker did not return consumer tag or it timed out? Can it be that sentry shows only last deprecation error and root cause is not visible? Ignoring empty `$consumer_tag` when `$nowait = false` might create another kind of issue..

janedbal:
`$nowait` was false, I assume it was caused by `AMQPNoDataException` in the wait method (possibly caused by SIGALRM).

Issue #1204, opened 2024-12-16
Zombie processes because of `SIGHeartbeatSender::unregister`


mrtus:
**Version(s) affected**: N/A
**PHP version**: N/A
**RabbitMQ version**: N/A

**Description**  
When using a heartbeat sender, `posix_kill` is used to kill the child process.
However this results in zombie processes that stay active because `pcntl_waitpid` is not called.

While maybe `SIGHeartbeatSender` is not meant to be ran in a PHP-FPM process, it sometimes has a use, we have FPM processes that can take some time, longer than the configured heartbeats, meaning the connection is always broken before a message is pushed on the connection

**How to reproduce**  
Run a PHP-FPM process that starts a heartbeat sender on every request.
Call `ps aux` and you'll see `php-fpx [defunct]` processes (aka zombie processes)

**Possible Solution**  
Call `pcntl_waitpid` after `posix_kill` is called;

https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Connection/Heartbeat/SIGHeartbeatSender.php#L58

Let me know your thoughts on this topic!
Thanks

ramunasd:
is method `unregister()` called before shutting down php-fpm process? This heartbeat sender does not ensure clean shutdown automatically.

mrtus:
> is method `unregister()` called before shutting down php-fpm process? This heartbeat sender does not ensure clean shutdown automatically.

Correct! I forgot to mention. The `unregister` method is called in a `register_shutdown_function` and I can confirm this is being executed. `SIGHeartbeatSender::unregister` contains the `posix_kill`.

ramunasd:
Can you confirm that `pcntl_waitpid()` inside shutdown function works as expected? I think it should be possible to use more general `pcntl_wait()` if cihld pid is not known.

mrtus:
> Can you confirm that `pcntl_waitpid()` inside shutdown function works as expected? I think it should be possible to use more general `pcntl_wait()` if child pid is not known.

We are using `pcntl_waitpid($childId)` in production for our own, similar heartbeat sender but for PDO, and it behaves as expected as there are no longer any zombie processes after each `unregister`.

I also tested this on the AMQP heartbeat sender and I confirm that `pcntl_waitpid($childId)` also there resolves the problem of zombie processes after `unregister` is called.

ramunasd:
Possible fix in PR https://github.com/php-amqplib/php-amqplib/pull/1205

[...]

PR #1244, opened 2026-09-16
Exit the SIGHeartbeatSender child process when it is orphaned

mrtus:
Hi 👋 

Let me know if you want me to change something.

This orphaned child is actively causing problems for us. So I thought I try to fix it "myself".

### Description

`SIGHeartbeatSender::periodicAlarm()` forks a child whose only way out is a `sleep()` interrupted by a signal. Nothing ever signals an orphan, so when the parent dies without executing any further PHP — `SIGKILL` among them — the child loops for the lifetime of the machine, re-parented to pid 1.

`unregister()` does not cover this, because it runs in the parent. This is the complement of #1204, which was about the zombie left behind when `unregister()` *does* run; the `pcntl_waitpid()` added for that case is unaffected here.

Beyond the leaked process, the orphan keeps signalling a pid that no longer exists, and pid numbers are recycled — so it can eventually deliver its signal to an unrelated process. That is why the check sits after the `sleep()` rather than at the top of the loop: at the top, an orphan still sends one last signal before noticing. It exits with `die` to match the bail-out above it, because breaking out of the loop would return the forked child into the caller's stack.

`PCNTLHeartbeatSender` has no equivalent failure mode, since it does not fork. No new extension requirement is introduced — `posix_kill()` is already called on this path.

The test is a unit test rather than a functional one because the child never touches the connection, so it is reachable through a mocked `AbstractConnection` with no broker running.

### Manual check

- [ ] Run a CLI script that opens a connection with a heartbeat, registers `SIGHeartbeatSender`, and sleeps. `kill -9` it, then `ps -o pid,ppid,command` → the child is gone within one heartbeat interval. On `master` it survives with ppid 1.
- [ ] Same script, but let it exit normally through `unregister()` → the child is still killed and reaped immediately, with no `[defunct]` entry left behind.


ramunasd:
Thanks a lot for your improvement. This is really important for stability and efficient resource usage.

I need to double check things and fix other tests before I can merge this one.

BTW, how do found info about `posix_getppid()`? Official documentation does not mention anything about missing parent process...

mrtus:
> Thanks a lot for your improvement. This is really important for stability and efficient resource usage.
> 
> I need to double check things and fix other tests before I can merge this one.
> 
> BTW, how do found info about `posix_getppid()`? Official documentation does not mention anything about missing parent process...

Thanks for the reply!

So we've had some issues on our own implementation of a `doctrine/dbal` variant of this heartbeat sender to keep our MySQL connection open in long-running processes. So this example of AMQP really helped for that!

We discovered that in some cases we ended up with orphaned processes when the heartbeat was running in a php-fpm or a-like environment that does not indicate that the process crashed/quit/... (e.g. [roadrunner](https://github.com/roadrunner-server/roadrunner) worker loops), this resulted in numerous orphans and huge memory leaks and broken file descriptors.

So we went on to investigate..., most of it is just Claude really helping us out on these edge-cases that require more in depth knowledge of these little systems 😁. (So to answer your question: Claude 🫣 is how we found out)

What I can confirm already is that this does work, it has been running on our "dbal" variant for about a month now, and over the past night it has been running for AMQP with my changes here and we are no longer seeing orphans after this change.

I've included the analysis of what we concluded in the past that explains everything a bit more in detail.

But this is the TL;DR

> The **child now compares `posix_getppid()` against the captured pid** after each sleep and immediately before each signal. That is what makes it safe under pid reuse rather than merely tidier: **once the real parent is gone the child is reparented to init, so `getppid()` returns 1** and can never again match a recycled number 

<details><summary>Analysis summary</summary>
<p>


`SIGHeartbeatSender::periodicAlarm()` forks a child that loops forever sending the sender's signal to a pid it captured once, before the fork. Both of that child's failure modes only bite when its parent dies, and both are fixed here. The design is deliberately unchanged — the fork stays, and so does the per-sender signal, because a process gets one `pcntl_alarm()` and one handler slot per signal, and consumers already run this sender in the same process as `php-amqplib`'s own `SIGHeartbeatSender` and `symfony/console`'s `SignalRegistry`.

The child inherited its parent's `stdout` and `stderr` and never released them. Under php-fpm those two descriptors are the write ends of the master's `catch_workers_output` pipes, and a pipe only reaches EOF once every write end is closed — so while the child lives, a dead worker's pipe never EOFs, the master never tears down the read event for it, and the next worker it forks is handed the same descriptor numbers back by `pipe()`. The master gives up on that collision with `unable to read what child N said into stdout: Bad file descriptor (9)`, after which every log write from the worker returns `EPIPE` and Monolog raises `UnexpectedValueException: Writing to the log file failed` for each one. The child writes nothing, so it closes both immediately after the fork; the parent keeps its own copies, so nothing it logs is affected. Descriptor 0 is left alone deliberately: only `stdout` and `stderr` are relayed by `catch_workers_output`, so it plays no part in this failure, and whatever fpm does point a worker's stdin at is not this class's business to close.

Nothing stopped the child when its parent died. `stop()` was the only thing that ever killed it, and `stop()` is reached through `register_shutdown_function`, which an abnormal exit skips — so a worker killed by a signal leaves the child behind, still firing every interval at a pid the kernel is free to reissue. php-fpm reissues worker pids briskly, so the stray signal lands on a freshly forked worker that has not installed a handler for it yet, and that worker dies on the signal's default disposition. In production this reads as `child 24 exited on signal 12 (SIGUSR2) after 4886 seconds`, its replacement starting in the same second, and the `EBADF` above arriving once that replacement has served enough traffic to log.

The child now compares `posix_getppid()` against the captured pid after each sleep and immediately before each signal. That is what makes it safe under pid reuse rather than merely tidier: once the real parent is gone the child is reparented to init, so `getppid()` returns 1 and can never again match a recycled number — the question being asked is about actual parentage, not about whether some process still answers to that pid. The check sits after the sleep rather than on the way into the loop because the parent dying mid-sleep is the entire scenario.

It terminates with `SIGKILL` to itself rather than `exit()`. The child is a fork of its parent's process image, so it carries every shutdown function the consuming application registered — including the `register_shutdown_function($sender->stop(...))` that wraps this class, and whatever closes the database connections it holds inherited handles to. It has never run any of them, and `exit()` would run all of them from a process that only ever slept and signalled. The previous code never ran them either, because the child never left the loop, so this preserves that property.

Two smaller corrections ride along in the same method. The `getmypid()` guard sat below the fork, where it read as the fork's guard, and now sits next to the call it narrows. And `pcntl_fork()` reports failure with `-1` rather than `false`, so nothing caught it: `-1` fell through the parent branch and was recorded as the child's pid, leaving a sender that looks started, has no child to signal it, and never fires a heartbeat. `stop()` tolerated that by testing `childPid > 0`, which is why it was survivable rather than visible.

The orphan case is covered by a new test that forks a parent, lets it start a sender, kills that parent without reaching `stop()`, and asserts the child is gone. Two things are **not** covered and should not be read as though they were: the `pcntl_fork() === -1` guard has no test, because the fork cannot be made to fail without exhausting the process table; and the `stdout`/`stderr` close has no test, because there is no php-fpm master in the suite and a forked child's descriptors are not observable from the process that forked it. For that second one, the pre-existing tests continuing to pass is the whole of the evidence that the heartbeat still works with both streams gone.


</p>
</details> 


Does this explain it for you @ramunasd ? Anything that I can help with?

mrtus:
Hi @ramunasd 

Looks like @syedusmanali already took the liberty to propose a fix for those failing tests https://github.com/php-amqplib/php-amqplib/pull/1242

Is anything blocking that merge?

How can I help to speed up the merge of these & have this important, and rather blocking (for us @teamleadercrm) , bugfix released?

ramunasd:
Thanks, one concern with this change: it relies on undocumented feature or side effect. That's a bit risky because older or future PHP versions might have different behavior. I don't rule out the possibility that function directly depends on posix extension version or even Linux kernel version.

Could you add functional test that use actual posix function? so that if underlying behavior ever shifts, the test goes red and we get an early signal.

mrtus:
> Thanks, one concern with this change: it relies on undocumented feature or side effect. That's a bit risky because older or future PHP versions might have different behavior. I don't rule out the possibility that function directly depends on posix extension version or even Linux kernel version.

Valid concern @ramunasd, and you are right, some documentation would be great on this.

I did some looking up, and this document https://man7.org/linux/man-pages/man2/getppid.2.html, since `posix_getppid` is essentially the same as `getppid()`, it describes its behaviour:

> getppid() returns the process ID of the parent of the calling process.  This will be either the ID of the process that created this process using fork(), or, if that process has already terminated, the ID of the process to which this process has been reparented (either init(1) or a "subreaper" process defined via the prctl(2) PR_SET_CHILD_SUBREAPER operation).

Which makes sense in every scenario I think? A process can never be bound to a process that does not exist and is always reparented when orphaned to a different "parent" (which has is Linux behaviour, as described on the man-pages).

This makes `posix_getppid() !== $parent` work in each scenario and should clear the concerns I think? Would you agree?

The page here about the exit function https://pubs.opengroup.org/onlinepubs/9799919799/functions/_exit.html also describes this behaviour of reparenting orphaned processes.

> The parent process ID of all of the existing child processes and zombie processes of the calling process shall be set to the process ID of an implementation-defined system process. That is, these processes shall be inherited by a special system process.

---

> Could you add functional test that use actual posix function? so that if underlying behavior ever shifts, the test goes red and we get an early signal.

Isn't that exact test already there with `child_exits_when_its_parent_is_gone`, which is a Functional test implementation? Is that what you meant, or what would you do differently?

---

Thank you for looking into this and making sure we cover our bases 👌

